PR #5, opened 2020-06-23
Added barebones for the profile page.

apluscs:


src/main/webapp/profile.html
```diff
@@ -0,0 +1,63 @@
...
+    <script src="profile.js"></script>
+  </head>
+  <body>
+    <div id="content">
+      <!-- TODO: add navbar here. -->
+      <h1>Your Profile</h1>
+      <form>
+        <!-- TODO: fill in values based on user data from Datastore. -->
+        <div class="form-group row">
+          <label for="name" class="col-sm-2 col-form-label">Name</label>
```
meagancheese, line 24: col-sm-2 col-form-label is not a super descriptive name for the class. Unsure what the col-sm-2 is.

apluscs: These are classes bootstrap uses to apply formatting to any element with that class. We need to leave them for formatting.

src/main/webapp/profile.html
```diff
@@ -0,0 +1,63 @@
...
+          <div class="col-sm-10">
```
meagancheese, line 25: Similar issue, not sure what col-sm-10 means

src/main/webapp/profile.html
```diff
@@ -0,0 +1,63 @@
...
+            <input type="text" readonly class="form-control-plaintext" id="user-name" value="Tess Ter">
+          </div>
+        </div>
+        <div class="form-group row">
+          <label for="userEmail" class="col-sm-2 col-form-label">Email</label>
+          <div class="col-sm-10">
+            <input type="text" readonly class="form-control-plaintext" id="user-email" value="tess@gmail.com">
```
meagancheese, line 32: This should not be editable, based on Google account they use to login.

apluscs: The readonly prevents it from being editable.

meagancheese (changes requested):
Some naming issues and I don't think that the user's email should be adjustable.

src/main/webapp/profile.html
meagancheese: Then why is it an input?

src/main/webapp/profile.html
```diff
@@ -0,0 +1,63 @@
...
+  <body>
+    <div id="content">
+      <!-- TODO: add navbar here. -->
+      <h1>Your Profile</h1>
+      <form>
+        <!-- TODO: fill in values based on user data from Datastore. -->
+        <div class="form-group row">
+          <label for="name" class="col-sm-2 col-form-label">Name</label>
+          <div class="col-sm-10">
+            <input type="text" readonly class="form-control-plaintext" id="user-name" value="Tess Ter">
```
meagancheese: This can be editable. People change their names from time to time.

src/main/webapp/profile.html
apluscs: Same as above.

src/main/webapp/profile.html
apluscs: l think it's so that it gets sent when the form gets submitted. l think it should be OK since it's from Bootstrap's docs?

meagancheese (approved):
LGTM

meagancheese:
If you want to attach a screenshot, do it as a comment. The pngs do not show up in GitHub.

acmcarther (approved):

PR #6, opened 2020-06-23
Splash/Home page

meagancheese:


src/main/webapp/index.html
```diff
@@ -16,7 +16,18 @@
...
     <div id="content">
-
+      <!--TODO: Add a navbar-->
+      <p id="welcome">Welcome to InterviewMe!</p>
+      <p>InterviewMe is a platform for conducting mock technical interviews where interviewers and interviewees alike can 
+        practice their skills. We aim to connect users with live video chats in order to encapsulate the true feeling
+        and pace of a technical interview.
+      </p>
+      <br>
+      <p>If you would like to logout, please click here.</p>
```
apluscs: l think we should incorporate login/logout as one portion of the navbar. Delete this?

target/maven-archiver/pom.properties
```diff
@@ -0,0 +1,9 @@
+#Generated by Maven
+<<<<<<< HEAD
+#Tue Jun 23 15:32:43 UTC 2020
+=======
+#Tue Jun 23 15:42:06 UTC 2020
+>>>>>>> 1b64b618d0eb3b78635f628ae690150fb59b9173
```
apluscs: Seems like a conflict to address.

src/main/webapp/style.css
```diff
@@ -1,5 +1,11 @@
...
-    width: 650px;
+  margin-left: auto;
+  margin-right: auto;
+  width: 650px;
+}
+
+#welcome {
+  font-size: 18;
+  text-align: center;
+  margin-bottom: 10px;
```
apluscs: Not a fan of specifying font sizes myself. l prefer using h1, h2, etc. to denote different font sizes. lt's more elegant that way and easier to track font sizes. l think we should try to Bootstrap as much as we can in general.
https://getbootstrap.com/docs/4.0/content/typography/

src/main/webapp/index.html
```diff
@@ -16,7 +16,18 @@
   </head>
   <body>
     <div id="content">
-
+      <!--TODO: Add a navbar-->
+      <p id="welcome">Welcome to InterviewMe!</p>
```
apluscs: change id to something more specific? like "welcome-heading"?

apluscs (changes requested):
l think it'd be great if we could get into the habit of attaching screenshots, especially when large changes are made to a page. Attach one please?

src/main/webapp/index.html
meagancheese: Done

target/maven-archiver/pom.properties
meagancheese: Done

src/main/webapp/index.html
meagancheese: Done

src/main/webapp/style.css
meagancheese: Done

meagancheese:
Good call, here you go: 
Screenshot 2020-06-23 at 5 21 18 PM

apluscs (changes requested):
For the Home Page, l think it looks better if the text is centered. Also the title is too big in comparison to the subtitle. Make the title smaller and the subtitle a larger? 
l usually model pages that need to be aesthetic after something professional like the ones here: https://line25.com/articles/showcase-of-web-designs-with-modern-splash-pages
Can't go wrong there!

apluscs (approved):
Love the look! LGTM.

src/main/webapp/index.html
```diff
@@ -16,7 +16,31 @@
   </head>
   <body>
     <div id="content">
-
+      <nav class="navbar navbar-expand-lg navbar-light bg-light">
+        <span class="navbar-brand mb-0 h1">InterviewMe</span>
+        <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarNavAltMarkup" aria-controls="navbarNavAltMarkup" aria-expanded="false" aria-label="Toggle navigation">
```
jalexanderqed: You should split this <button> tag, possibly twice, so that it is not overly long.

jalexanderqed (approved):
One change requested, but otherwise LGTM

src/main/webapp/index.html
meagancheese: Done

PR #7, opened 2020-06-23
Added hardcoded Scheduled Interviews page

apluscs:


meagancheese (approved):
LGTM

meagancheese:
Attach screenshots as a comment or in the PR description. We can't see them in GitHub when it is just a file in the repo.

src/main/webapp/scheduled-interviews.html
```diff
@@ -0,0 +1,55 @@
...
+      <section id="scheduled-interviews-cards">
+        <div class="row">
+          <div class="col-sm-6">
+            <div class="card" style="width: 18rem;">
+              <div class="card-body">
+                <h5 class="card-title">July 17, 2020</h5>
+                <p class="card-text">3:00PM - 4:00PM EST</p>
+              </div>
+              <ul class="list-group list-group-flush">
+                <li class="list-group-item">Your role: Interviewer</li>
```
jalexanderqed: I think the user's role should be more prominent, probably at the top of the card with the same size as the date (<h5>). You may want to reformat it as "You are interviewing <first name of interviewee>" and "<first name of interviewer> is interviewing you," but the specifics of that are up to you. The main point is that it should be easy to tell at a glance what the user's role will be.

jalexanderqed:
@meagancheese It's possible to view the screenshot by clicking the "view rich diff" button on the right next to the file name.

jalexanderqed (changes requested):

meagancheese (changes requested):
Is there an easy way to make it so that the card is not much bigger than the text?

meagancheese (changes requested):
Quick fix, instead of creating the scheduled-interviews-heading id you can apply the class "text-center"

meagancheese (approved):
LGTM

jalexanderqed (approved):

PR #12, opened 2020-06-24
Availability Page

meagancheese:


src/main/webapp/index.html
```diff
@@ -29,7 +29,7 @@
             <!--TODO: Put in links to other pages. Put on all pages-->
             <a class="nav-item nav-link" href="#">Profile</a>
             <a class="nav-item nav-link" href="#">Events</a>
-            <a class="nav-item nav-link" href="#">Availability</a>
+            <a class="nav-item nav-link" href="availability.html">Availability</a>
```
apluscs, line 32: Does this work? lt doesn't need the slash in front?

meagancheese: It works, yeah.

PR #16, opened 2020-06-24
Adds navbars to pages on master

meagancheese:


apluscs (approved):
lt might be better to leave this until we get all the HTML pages into master. lt'll defintely save us time (we can just copy one navbar to all pages) and it'll be easier to track right now.
l don't really see the point of doing this right now. l'll approve but we don't have to do this for other pages yet.

src/main/webapp/profile.html
```diff
@@ -16,8 +16,28 @@
   </head>
   <body>
     <div id="content">
-      <!-- TODO: add navbar here. -->
-      <h1>Your Profile</h1>
+      <nav class="navbar navbar-expand-lg navbar-light bg-light">
```
acmcarther, line 19: Duplicating this bar on every page is not going to be maintainable, because you will have to update this in lock step in all pages.

Not a problem in the prototype stage, but make a mental note to look into a templating engine for the real thing. A good templating engine with builtin support in app engine is JSP.

Here's the relevant documentation: https://cloud.google.com/appengine/docs/standard/java/config/webxml#jsps There are also tutorials available.

acmcarther (approved):
cc @jalexanderqed, I've recommended JSP as the templating solution in this PR. Looks like it has support in AppEngine.

PR #18, opened 2020-06-25
Search and selection of interview for interviewee. With confirmation.

apluscs:
1. search-1
2. search-2
3. search-3
4. search-4


src/main/webapp/search-interview.html
```diff
@@ -0,0 +1,107 @@
...
+      <section hidden id="search-results">
+        <!--TODO: Include Timezone consideration; Add pagination; Render these interview times dynamically;-->
+        <form>
+          <div class="form-row">
+            <div class="col-3 date-label">
+              <label>Monday 6/29</label>
+            </div>
+            <div class="col-5">
+              <select class="form-control">
+                <option>8:15 AM - 9:15 PM</option>
```
meagancheese: Should be 8:15 AM - 9:15 AM

src/main/webapp/search-interview.html
```diff
@@ -0,0 +1,107 @@
...
+            <div class="col-4">
+              <button type="button" class="btn btn-primary mb-2">Select</button>
+            </div>
+          </div>
+        </form>
+        <br>
+        <form>
+          <div class="form-row">
+            <div class="col-3 date-label">
+              <label>Wednesday 6/30</label>
```
meagancheese: Wednesday is 7/1

src/main/webapp/search-interview.html
```diff
@@ -0,0 +1,107 @@
...
+            <div class="col-4">
+              <button type="button" class="btn btn-primary mb-2">Select</button>
+            </div>
+          </div>
+        </form>
+        <br>     
+        <form>
+          <div class="form-row">
+            <div class="col-3 date-label">
+              <label>Thursday 7/1</label>
```
meagancheese: Thursday is 7/2

src/main/webapp/search-interview.html
```diff
@@ -0,0 +1,107 @@
...
+            <div class="col-4">
+              <button type="button" class="btn btn-primary mb-2">Select</button>
+            </div>
+          </div>
+        </form>
+        <br>  
+        <form>
+          <div class="form-row">
+            <div class="col-3 date-label">
+              <label>Sunday 7/1</label>
```
meagancheese: Are we doing Sundays? If so, our week should be Sun-Saturday. Also check the date. Might be easiest to switch to Saturday 7/4

src/main/webapp/search-interview.html
```diff
@@ -0,0 +1,107 @@
...
+            </div>
+            <div class="col-5">
+              <select class="form-control">
+                <option>9:30 AM - 10:30 AM</option>
+                <option>9:30 AM - 10:30 AM</option>
```
meagancheese: We should not have duplicate times. We pick the interviewer for them.

meagancheese (changes requested):
Some small changes, otherwise looks good. Also, put the pictures in the PR description, not in the files themselves.

src/main/webapp/search-interview.html
apluscs: Ok we can do Sun-Saturday and l'll change to Saturday.

apluscs:
1. search-1
2. search-2
3. search-3
4. search-4

meagancheese (changes requested):
Delete the screenshots from the file and put them in the PR description instead (at the top, not in the comments).

meagancheese (approved):
LGTM

src/main/webapp/search-interview.html
```diff
@@ -0,0 +1,106 @@
...
+      </form>
+      <br>      
+      <section hidden id="search-results">
+        <!--TODO: Include Timezone consideration; Add pagination; Render these interview times dynamically;-->
+        <form>
+          <div class="form-row">
+            <div class="col-3 date-label">
+              <label>Monday 6/29</label>
+            </div>
+            <div class="col-5">
```
jalexanderqed: It's not immediately clear from looking at the page (as posted in the description) that these are the available interview times. It would be helpful to have a header at the top of the list of times with something like "Available Interviews".

src/main/webapp/search-interview.html
```diff
@@ -0,0 +1,106 @@
...
+              <select class="form-control">
+                <option>8:15 AM - 9:15 AM</option>
+                <option>10:30 AM - 11:30 AM</option>
+                <option>1:00 PM - 2:00 PM</option>
```
jalexanderqed, line 46: Just a note: this display method is fine for now, but I think a more helpful and intuitive one would be to show all the interview times in a calendar grid, like the weekly view in Google Calendar. This will be a more complex interface, so you don't have to change it now, but if it's something you would like to do later (I think it would be worth the time), please add a TODO.

jalexanderqed (changes requested):

src/main/webapp/search-interview.html
meagancheese: One of the problems I see with this is that we will have overlapping events, since we are showing them all possible times. Worst/best-case scenario an event every 15 minutes.

jalexanderqed: That's true but still seems like a solvable problem with some inventive interface design. Definitely more complex than I originally envisioned though.

src/main/webapp/search-interview.html
apluscs: Done. Updated the PR description.

src/main/webapp/search-interview.html
apluscs: Still don't see how the calendar would have the overlapping events problem, unless we put in lines for the start times instead of blocks... l put the TODO in though, so we can discuss and decide later.

jalexanderqed (approved):

PR #20, opened 2020-06-25
Static Registration Page 

apluscs:
Straight from login in this is what you see: register-1
When you don't enter a required field: register-2


src/main/webapp/register.html
```diff
@@ -0,0 +1,100 @@
...
+            <input type="text" readonly class="form-control" id="user-email" value="tess@gmail.com">
+          </div>
+        </div>
+        <div class="form-row">
+          <div class="col-md-6 mb-3">
+            <label for="first-name-field">First name</label>
+            <input type="text" class="form-control" id="first-name-field" placeholder="First name" value="" required>
+            <div class="valid-feedback">
+              Looks good!
+            </div>
```
meagancheese: I think the checks are enough

src/main/webapp/register.html
```diff
@@ -0,0 +1,100 @@
...
+          </div>
+          <div class="col-md-6 mb-3">
+            <label for="first-name-field">Last name</label>
+            <input type="text" class="form-control" id="last-name-field" placeholder="Last name" value="" required>
+            <div class="valid-feedback">
+              Looks good!
+            </div>
```
meagancheese: Same here

src/main/webapp/register.html
```diff
@@ -0,0 +1,100 @@
...
+          </div>
+        </div>
+        <div class="form-row">
+          <div class="col-md-6 mb-3">
+            <label for="company-field">Company</label>
+            <input type="text" class="form-control" id="company-field" placeholder="Google" value="">
+          </div>
```
meagancheese, line 68: Is there a reason this one didn't get a check mark?

src/main/webapp/register.html
```diff
@@ -0,0 +1,100 @@
...
+        <button class="navbar-toggler" type="button" data-toggle="collapse" 
+          data-target="#navbarNavAltMarkup" aria-controls="navbarNavAltMarkup" 
+          aria-expanded="false" aria-label="Toggle navigation">
+        <span class="navbar-toggler-icon"></span>
+        </button>
+        <div class="collapse navbar-collapse" id="navbarNavAltMarkup">
+          <div class="navbar-nav">
+            <a class="nav-item nav-link active" href="#">Home</a>
+            <!--TODO: Put in links to other pages. Put on all pages-->
+            <a class="nav-item nav-link" href="profile.html">Profile</a>
```
meagancheese, line 32: Which should be active here? I'm thinking Profile. Also, none of the links should work while someone is registering. Either address this now or leave a TODO. I'm working on it in a different branch for logged in/ logged out.

meagancheese (changes requested):
Some details

src/main/webapp/register.html
apluscs: l think we should just have the Home page as an option and remove everything else. 
For the home page itself, if they aren't logged in, it should just be home and log in as options.
What do you think?

meagancheese: This is where they're directed after they log in though and they can't do anything until they're registered. I'd prefer to have profile highlighted as active and all other links redirect to the current page, until the user submits valid information.

meagancheese: Also, separate Q: Are the job, company, and linkedin slots mandatory? If they are currently, I don't think they should be. Think about an interviewee using the site.

src/main/webapp/register.html
apluscs: Seems like this only happens when you click a suggestion. The background image (check mark) doesn't show. But when you type it in, it's fine. l'll update the screenshot.

meagancheese (changes requested):
Add a TODO about restricting user access to other pages.

meagancheese (approved):
LGTM

jalexanderqed (approved):

meagancheese:
Purposefully not merging rn, will after prototype is presented.

PR #22, opened 2020-06-25
RegistrationServlet to send newly registered user info to Datastore

apluscs:


meagancheese (approved):
LGTM. Note for future, we still have to get the user's email and auto-fill it on this form.

PR #24, opened 2020-06-26
updated form on profile page to match the one on the registration page

apluscs:
1. profile-1
2. profile-2


src/main/webapp/profile.html
```diff
@@ -14,8 +14,9 @@
     <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"
       integrity="sha384-wfSDF2E50Y2D1uUdj0O3uMBJnjuUD4Ih7YwaYd1iqfktj0Uod8GCExl3Og8ifwB6" crossorigin="anonymous"></script>
     <script src="profile.js"></script>
+    <script src="register.js"></script>    
   </head>
-  <body>
+  <body onload="onloadProfile()">
```
meagancheese: In another branch I was doing onProfileLoad()

src/main/webapp/profile.html
```diff
@@ -39,36 +40,48 @@
...
-          <label for="user-job" class="col-sm-2 col-form-label">Job</label>
-          <div class="col-sm-10">
-            <input type="text" readonly class="form-control-plaintext editable" id="user-job" value="Software Engineer">
+        <div class="form-row">
+          <div class="col-md-6 mb-3">
+            <label for="company-field">Company</label>
+            <input type="text" class="form-control-plaintext editable" id="company-field" value="Google">
+          </div>
+          <div class="form-group col-md-6 mb-3">
+            <label for="job-field">Job</label>
```
meagancheese, line 67: Why is this showing up as a dropdown even when it's set?

src/main/webapp/profile.js
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+function onloadProfile() {
```
meagancheese: Again, switch to onProfileLoad()

meagancheese (changes requested):
Small changes, one for conformity, the other for aesthetics.

src/main/webapp/profile.html
apluscs: Software Engineer is the first option

meagancheese: When we are showing the user the current info they inputted, we shouldn't have the job entry look like it's still modifiable.

meagancheese (changes requested):
See comment about Job display.

src/main/webapp/profile.html
apluscs: Updated screenshots.

meagancheese (approved):
LGTM

src/main/webapp/profile.html
```diff
@@ -39,36 +40,49 @@
...
-          <label for="user-company" class="col-sm-2 col-form-label">Company</label>
-          <div class="col-sm-10">
-            <input type="text" readonly class="form-control-plaintext editable" id="user-company" value="Google">
+        <div class="form-row">
+          <div class="col-md-6 mb-3">
+            <label for="first-name-field">First name</label>
+            <input type="text" readonly class="form-control-plaintext editable" id="first-name-field" placeholder="First name" value="Tess" required>
+          </div>
+          <div class="col-md-6 mb-3">
+            <label for="first-name-field">Last name</label>
```
jalexanderqed: It seems like this should be `for="last-name-field"`

jalexanderqed (approved):
One small fix requested. Approving now.

PR #27, opened 2020-06-26
Disables navbar links while on registration page.

meagancheese:
Screenshot 2020-06-26 at 4 48 44 PM

### DIFF
--- a/src/main/webapp/register.html
+++ b/src/main/webapp/register.html
@@ -38,7 +38,6 @@
             <a class="nav-item nav-link disabled" href="search-interviews.html">Scheduling</a>
             <!--TODO: Move this to the right of the navbar-->
             <a class="nav-item nav-link" href="#" id="login-tab">Logout</a>
-            <!--CHANGE TO TEST NEW CLOUDBUILD SETTINGS-->
           </div>
         </div>
       </nav>
--- a/src/main/webapp/register.html
+++ b/src/main/webapp/register.html
@@ -38,6 +38,7 @@
             <a class="nav-item nav-link disabled" href="search-interviews.html">Scheduling</a>
             <!--TODO: Move this to the right of the navbar-->
             <a class="nav-item nav-link" href="#" id="login-tab">Logout</a>
+            <!--CHANGE TO TEST NEW CLOUDBUILD SETTINGS-->
           </div>
         </div>
       </nav>
--- a/src/main/webapp/register.html
+++ b/src/main/webapp/register.html
@@ -14,8 +14,11 @@
     <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js"
       integrity="sha384-wfSDF2E50Y2D1uUdj0O3uMBJnjuUD4Ih7YwaYd1iqfktj0Uod8GCExl3Og8ifwB6" crossorigin="anonymous"></script>
     <script src="register.js"></script>
+    <script src="logout.js"></script>
   </head>
-  <body onload="prepareFormValidation()">
+  <body onload="onRegisterLoad()">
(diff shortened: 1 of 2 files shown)


src/main/webapp/register.html
```diff
@@ -26,15 +29,15 @@
...
-            <a class="nav-item nav-link active" href="#">Home</a>
-            <!--TODO: Put in links to other pages. Put on all pages-->
-            <a class="nav-item nav-link" href="profile.html">Profile</a>
-            <a class="nav-item nav-link" href="scheduled-interviews.html">Events</a>
-            <a class="nav-item nav-link" href="availability.html">Availability</a>
-            <a class="nav-item nav-link" href="#">Scheduling</a>
-            <!--TODO: Move this to the right of the navbar and allow user to log in and out using Authentication-->
-            <a class="nav-item nav-link" href="#">Logout</a>
+            <!--TODO: Redirect user to this page if they attempt to go to another. Will depend on being able to tell
+              if a user is registered or not.-->
```
apluscs, line 33: Shouldn't it depend on whether or not the user is logged in?

meagancheese: No, if they're not logged in they'll still be redirected to home but the case here is they're logged in but not registered (no user w/ their email is in datastore).

acmcarther:
This (and other) PRs would benefit from terse explanations in the PR description.

If I understand your intent, you're trying to avoid people clicking out of the registration page while still filling the form out. There's a more conventional mechanism for this:

https://stackoverflow.com/questions/7317273/warn-user-before-leaving-web-page-with-unsaved-changes

meagancheese:
It goes farther than not letting them click out of the page while filling out the form. We want it to be so that the user is redirected to the registration page when it is their first time logging in (so we don't have a User object for their email yet) and then make it so that they are not allowed to go to any other page on the site until they have registered. If they type in another page's url, they should be reidrected back to the registration page. Also, the link that you provided still gives them the option of going to another page if they say they're sure they want to go.

src/main/webapp/register.js
```diff
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+function onRegisterLoad() {
+  supplyLogoutLink();
+  prepareFormValidation();
+}
+
+
```
apluscs: Remove this line

apluscs (approved):
One nit, but LGTM

src/main/webapp/register.html
```diff
@@ -26,15 +29,15 @@
...
-            <!--TODO: Put in links to other pages. Put on all pages-->
-            <a class="nav-item nav-link" href="profile.html">Profile</a>
-            <a class="nav-item nav-link" href="scheduled-interviews.html">Events</a>
-            <a class="nav-item nav-link" href="availability.html">Availability</a>
-            <a class="nav-item nav-link" href="#">Scheduling</a>
-            <!--TODO: Move this to the right of the navbar and allow user to log in and out using Authentication-->
-            <a class="nav-item nav-link" href="#">Logout</a>
+            <!--TODO: Redirect user to this page if they attempt to go to another. Will depend on being able to tell
+              if a user is registered or not.-->            
+            <a class="nav-item nav-link disabled" href="index.html">Home</a>
```
apluscs, line 34: Home is disabled? Shouldn't they have the opportunity to go back?

src/main/webapp/register.js
jalexanderqed: AGreed this looks like an extra line

jalexanderqed (approved):

src/main/webapp/register.js
meagancheese: Done

PR #39, opened 2020-06-30
Added DatastorePersonDao and PersonDAO interface; Tested put() with registration servlet

apluscs:
l know my Exception handling is trash, that'll go in another PR.


src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,77 @@
...
+    personEntity.setProperty("email", person.key());    
+    personEntity.setProperty("firstName", person.getFirstName());
+    personEntity.setProperty("lastName", person.getLastName());
+    personEntity.setProperty("company", person.getCompany());
+    personEntity.setProperty("job", person.getJob());
+    personEntity.setProperty("linkedIn", person.getLinkedIn());
+    return personEntity;
+  }
+
+  /* Since email is a personEntity's key, we retrieve the corresponding personEntity from Datastore and return it as a Person object. */
```
acmcarther: Line is too long

src/main/java/com/google/sps/data/PersonDao.java
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+/* PersonDao includes the basic methods anything managing Person entities must support */
+public interface PersonDao {
+  // Returns a Person object from storage, with email as the key.
+  public Person get(String email) throws com.google.appengine.api.datastore.EntityNotFoundException;
```
acmcarther: Yeah I think you're aware, but PersonDao shouldn't be throwing an exception specific to one implementation of the interface.

(That said, I've got a possible remedy for this as a comment in the other file)

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,77 @@
...
+  // TODO: make our own PersonNotFoundException and refactor how we're handling these
+  public Person get(String email)
+      throws com.google.appengine.api.datastore.EntityNotFoundException {
```
acmcarther: One option here is to return Optional<Person> from get.

To do that, catch EntityNotFoundException in the implementation and return Optional.empty().

(the type is java.util.Optional).

src/main/java/com/google/sps/data/PersonDao.java
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+/* PersonDao includes the basic methods anything managing Person entities must support */
```
acmcarther: Nit: period at end of sentence

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,77 @@
...
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+
+  /* We make an entity in Datastore with person's fields as properties and their email as the key. */
+  public void put(Person person) {
+    datastore.put(personToEntity(person));
+  }
+
+  private static Entity personToEntity(Person person) {
+    Entity personEntity = new Entity("Person", person.key());
```
GrantComm: Change .key() to .email() to reflect current changes in person class

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,77 @@
...
+    personEntity.setProperty("email", person.key());
```
GrantComm: here too

acmcarther (changes requested):
Generally looks good. I suggest that you return Optional from get so that the application explicitly has to check that the value exists.

GrantComm (changes requested):
Besides changing the name of the function to access the user email lgtm

src/main/java/com/google/sps/data/DatastorePersonDao.java
apluscs: lf l return an Optional, then higher up in the call chain (in the PersonServlet) l have to check if the Optional is empty and then do my business logic (redirect to registration page). ls this what you want? 
l was planning on throwing a PersonNotFoundException that l define and then that propagate up to PersonServlet

GrantComm (approved):
LGTM

src/main/webapp/register.html
```diff
@@ -51,27 +51,33 @@ <h5 class="text-center lead">Fill in the following fields below.</h5>
...
           </div>
         </div>
         <div class="form-row">
           <div class="col-md-6 mb-3">
             <label for="first-name-field">First name</label>
-            <input type="text" class="form-control" id="first-name-field" placeholder="First name" value="" required>
+            <input type="text" class="form-control" id="first-name-field" placeholder="First name" value="" required name="first-name">
+            <div class="valid-feedback">
+              Looks good!
+            </div>
```
meagancheese: Why did we put these back? Were the green check marks not enough?

src/main/webapp/register.html
```diff
@@ -51,27 +51,33 @@ <h5 class="text-center lead">Fill in the following fields below.</h5>
...
           </div>
           <div class="col-md-6 mb-3">
             <label for="first-name-field">Last name</label>
-            <input type="text" class="form-control" id="last-name-field" placeholder="Last name" value="" required>
+            <input type="text" class="form-control" id="last-name-field" placeholder="Last name" value="" required name="last-name">
+            <div class="valid-feedback">
+              Looks good!
+            </div>
```
meagancheese: Same thing here.

meagancheese (approved):
QQ, but otherwise LGTM.

src/main/webapp/register.html
apluscs: Ah it's because the other PR in which l changed these got closed.

src/main/java/com/google/sps/data/PersonDao.java
acmcarther: This hasn't been fixed.

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /** Initializes the fields for PersonDatastoreDAO. */
+  public DatastorePersonDao() {
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+
+  /**
+   * We make an entity in Datastore with person's fields as properties and their email as the key.
+   */
+  public void put(Person person) {
```
acmcarther, line 47: `@Override`

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,82 @@
...
+    personEntity.setProperty("linkedIn", person.getLinkedIn());
+    return personEntity;
+  }
+
+  /**
+   * Since email is a personEntity's key, we retrieve the corresponding personEntity from Datastore
+   * and return it as a Person object.
+   */
+  // TODO: make our own PersonNotFoundException and refactor how we're handling these
+  public Person get(String email)
```
acmcarther: `@Override`

src/main/java/com/google/sps/data/DatastorePersonDao.java
acmcarther: Having a dedicated exception for handling "personnotfound" doesn't buy you much. I think it's really only useful in this one context.

It could make sense to have an exception here, if you never expect the person to be missing. That said, I think Optional better expresses the semantics. If the caller cannot handle the absence of the value (i.e. it cannot render a servlet without it), then the Optional is exactly what you want, because it forces you to handle this case directly instead of just pushing the exception upward.

acmcarther (changes requested):
will wait on the Optional change.

src/main/java/com/google/sps/data/DatastorePersonDao.java
apluscs: l'll add it but what are we overriding? This isn't the exact method from Datastore.

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,88 @@
...
+   * and return it as a Person object.
+   */
+  @Override
+  public Optional<Person> get(String email) {
+    // TODO: handle this with the profile page servlet
+    Key key = KeyFactory.createKey("Person", email);
+    Entity personEntity;
+    try {
+      personEntity = datastore.get(key);
+    } catch (Exception e) {
```
acmcarther: Generally, you want to catch the most specific exception possible.

What you have here will catch basically any exception, including exceptions you really shouldn't be accidentally catching.

If you need to catch more than just "entity not found", consider rethrowing the other type of exception.

acmcarther (changes requested):
LGTM, but I want to see what exception you use for get().

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -0,0 +1,90 @@
...
+  @Override
+  public Optional<Person> get(String email) {
+    // TODO: handle this with the profile page servlet
+    Key key = KeyFactory.createKey("Person", email);
+    Entity personEntity;
+    try {
+      personEntity = datastore.get(key);
+    } catch (com.google.appengine.api.datastore.EntityNotFoundException e) {
+      return Optional.empty();
+    } catch (Exception e) {
```
acmcarther: Why are you catching this?

Readers will wonder what exceptions this is hiding.

(1) Do not catch very general exceptions, basically ever (unless you're writing a watchdog for a process)
(2) When you do, clearly document *what* is being caught and *why* you're catching it that way.

acmcarther (approved):
LGTM

PR #41, opened 2020-06-30
Gets a person's info from Datastore and autofills the profile page

apluscs:
Exception handling will go in another PR.


src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -46,7 +46,7 @@ public void put(Person person) {
 
   private static Entity personToEntity(Person person) {
     Entity personEntity = new Entity("Person", person.key());
-    personEntity.setProperty("email", person.key());    
+    personEntity.setProperty("email", person.key());
```
GrantComm: change key() to getEmail()

GrantComm (changes requested):
Other than methods used LGTM

PR #44, opened 2020-06-30
Skeleton for Scheduled interview class

GrantComm:


src/main/java/com/google/sps/data/ScheduledInterview.java
```diff
@@ -0,0 +1,44 @@
...
+  private String interviewerEmail;
+  private String intervieweeEmail;
+
+  public ScheduledInterview(
+      TimeRange when, String interviewerEmail, String intervieweeEmail) {
+    this.when = when;
+    this.interviewerEmail = interviewerEmail;
+    this.intervieweeEmail = intervieweeEmail;
+  }
+
```
apluscs, line 27: Still need setter methods.

src/main/java/com/google/sps/data/ScheduledInterview.java
```diff
@@ -0,0 +1,44 @@
...
+package com.google.sps;
+
+/** Represents a scheduled interview. Contains the timerange of the interview and the emails of the attendees. */
+public class ScheduledInterview {
+  private TimeRange when;
+  private String interviewerEmail;
+  private String intervieweeEmail;
+
+  public ScheduledInterview(
```
apluscs: Javadoc for this

src/main/java/com/google/sps/data/ScheduledInterview.java
```diff
@@ -0,0 +1,44 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+/** Represents a scheduled interview. Contains the timerange of the interview and the emails of the attendees. */
+public class ScheduledInterview {
+  private TimeRange when;
```
apluscs: Question: l see TimeRange doesn't exist yet, so this won't compile. Put a TODO here?

apluscs (changes requested):

apluscs (approved):

src/main/java/com/google/sps/data/ScheduledInterview.java
```diff
@@ -0,0 +1,73 @@
...
+  /** Sets the interviewer email. */
+  public void setInterviewerEmail(String email) {
+    interviewerEmail = email; 
+  }
+
+  /** Sets the interviewee email. */
+  public void setIntervieweeEmail(String email) {
+    intervieweeEmail = email; 
+  }
+  /** Returns the timerange of the interview */
```
meagancheese: Add space for consistency.

src/main/java/com/google/sps/data/ScheduledInterview.java
meagancheese: General question: what is the purpose of the setter methods? In what situations would we be updating these after the initialization? @GrantComm  @apluscs

meagancheese (changes requested):
Small change and general question.

meagancheese (approved):
LGTM

src/main/java/com/google/sps/data/ScheduledInterview.java
```diff
@@ -0,0 +1,80 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import java.time.LocalDate;
+
+/** Represents a scheduled interview. */
+public class ScheduledInterview {
```
acmcarther: AutoValue?

Copy-paste of my comment from https://github.com/googleinterns/InterviewMe/pull/43/files but updated for your use case

---

https://github.com/google/auto/tree/master/value

Here's what the current version of the code might look like with autovalue:
```java
  import com.google.auto.value.AutoValue;

  @AutoValue
  abstract static class ScheduledInterview {
    abstract TimeRange when();
    abstract LocalDate date();
    abstract String interviewerEmail();
    abstract String intervieweeEmail();

    static ScheduledInterview create(TimeRange when, LocalDate date, String interviewerEmail, String intervieweeEmail) {
      return new AutoValue_ScheduledInterview(when, date, interviewerEmail, intervieweeEmail);
    }
  }
```
I can give the pod a primer on AutoValue if would be helpful. I find it incredibly useful both to simplify code, and also as a tool for thinking about problems, since I find the best code strongly distinguishes between simple value types (i.e. above), and classes which do lots of work using value types (i.e. a servlet class).

src/main/java/com/google/sps/data/ScheduledInterview.java
acmcarther: Relevant here:

If you use autovalue, you don't get setters, though you can "re-build" the type by using AutoValue builders, and initializing a new value from the old value (but "setting" the new fields in the builder)

https://github.com/google/auto/blob/master/value/userguide/builders.md

acmcarther (changes requested):
Note on autovalue

src/main/java/com/google/sps/data/ScheduledInterview.java
```diff
@@ -0,0 +1,56 @@
...
+// limitations under the License.
+
+package com.google.sps.data;
+
+import com.google.auto.value.AutoValue;
+import java.time.LocalDate;
+
+/** Represents a scheduled interview. */
+@AutoValue
+abstract class ScheduledInterview {
```
acmcarther: Should this be public?

(sorry for being misleading in my example...)

acmcarther (approved):
LGTM % missing `public`s

PR #46, opened 2020-07-01
Scheduled interview dao interface

GrantComm:


src/main/java/com/google/sps/data/ScheduledInterviewDao.java
```diff
@@ -0,0 +1,27 @@
...
+
+package com.google.sps;
+
+/**
+ * ScheduledInterviewDao includes the basic methods anything managing ScheduledInterview entities
+ * must support.
+ */
+public interface ScheduledInterviewDao {
+  // Returns a ScheduledInterview object from storage
+  public ScheduledInterview get() throws com.google.appengine.api.datastore.EntityNotFoundException;
```
apluscs: TODO here about possibly changing this exception handling to an Optional or some exception we define. We'll see what happens in PR 39

apluscs (approved):
Just a TODO

meagancheese (approved):
LGTM

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
jalexanderqed: This probably needs to take the key of the ScheduledInterview that you want to retrieve. A more helpful interface might be to get all the ScheduledInterview records for a particular user, but special methods like that can be added later. For now, it definitely needs to take some kind of key.

jalexanderqed: I think we discussed yesterday that you should wrap the datastore-specific exception objects in your own class. Here are some instructions on doing that: https://www.baeldung.com/java-new-custom-exception. Your custom NotFound exception should probably take a com.google.appengine.api.datastore.EntityNotFoundException as a constructor argument.

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
```diff
@@ -0,0 +1,27 @@
...
+
+  // Puts a ScheduledInterview object into storage.
+  public void put(ScheduledInterview scheduledInterview);
```
jalexanderqed: Consider having create() and update() methods rather than just put. I think this generally keeps the logic cleaner both inside the methods and at the call sites. You don't necessarily have to do this, but I'd like to hear your reasoning if you don't.

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
jalexanderqed: I meant that you should also delete `put` here since it is no longer necessary with `create` and `delete`. Sorry, I see why this wasn't clear from my first comment.

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
```diff
@@ -0,0 +1,37 @@
...
+  public Optional<ScheduledInterview> get(String email);
+
+  // Puts a ScheduledInterview into storage.
+  public void put(ScheduledInterview scheduledInterview);
+
+  // Updates a ScheduledInterview in storage.
+  public void update(ScheduledInterview oldScheduledInterview);
+
+  // Creates a ScheduledInterview entity.
+  public Entity create(ScheduledInterview scheduledInterview);
```
jalexanderqed, line 33: Minor: `create()` usually comes before `update`.

GrantComm: Done

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
GrantComm: Did you mean create and update here?

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.google.appengine.api.datastore.Entity;
+import java.util.Optional;
+
+/**
+ * ScheduledInterviewDao includes the basic methods anything managing ScheduledInterview entities
+ * must support.
+ */
+public interface ScheduledInterviewDao {
+  // Returns a the most first ScheduledInterview object from storage using the users email as the
```
jalexanderqed: This comment seems grammatically problematic. (a the most first)

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
```diff
@@ -0,0 +1,36 @@
...
+  // key.
+  public Optional<ScheduledInterview> get(String email);
```
jalexanderqed: Is this actually a useful `get()` interface for scheduled interviews? It seems like returning all interviews for a particular user would be required in most cases.

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
jalexanderqed: `using the user's email` is not very specific as to which interviews will be returned. Is it all interviews in which the user will participate (as any participant)? If so, the email cannot be the key since both the interviewer and interviewee's email cannot be the key. `getForPerson()` might be a more appropriate method and would return all interviews in which a user will participate.

jalexanderqed (changes requested):
Sorry, I rushed through my previous review and didn't pay enough attention to get(). Please discuss with Claire and Meagan if necessary what behavior you expect to need from this DAO. Getting by key is actually unlikely to be useful for ScheduledInterview objects, though you should decide what key you are going to use (it might just be the auto-generated Datastore one).

src/main/java/com/google/sps/data/ScheduledInterviewDao.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Optional;
+import java.util.List;
+
+/**
+ * ScheduledInterviewDao includes the basic methods anything managing ScheduledInterview entities
+ * must support.
+ */
+public interface ScheduledInterviewDao {
+  // Returns the ScheduledInterview object.
+  public Optional<ScheduledInterview> get(Long id);
```
jalexanderqed: This should take `long` (the primitive type) rather than the class wrapper (java.lang.Long), which you have here.

jalexanderqed (changes requested):

jalexanderqed (approved):

PR #49, opened 2020-07-06
Autofills profile info and redirects to register if not registered yet.

apluscs:


src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -49,5 +50,23 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      if (!personOpt.isPresent()) {
+        // TODO: apply this to all other pages when someone accesses them illegally.
+        response.sendRedirect("/register.html");
+        return;
+      }
+      response.setContentType("application/json;");
+      response.getWriter().println(new Gson().toJson(personOpt.get()));
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
```
meagancheese: Avoid general Exceptions. Try and find the specific one(s) for this situation.

src/main/webapp/profile.js
```diff
@@ -13,15 +13,37 @@
 // limitations under the License.
 
 function onProfileLoad() {
+  autofillForm(getUserIfRegistered());  
+  supplyLogoutLink();
```
meagancheese: supplyLogoutLink() should be called first because it redirects the user to the home page if they are not logged in.

src/main/webapp/profile.js
```diff
@@ -13,15 +13,37 @@
...
-function onProfileLoad() {
-  supplyLogoutLink();
+// Fills in the profile form with data from Datastore.
+function autofillForm(person) {
+  document.getElementById("user-email").value = person.email;
+  document.getElementById("first-name-field").value = person.firstName;
+  document.getElementById("last-name-field").value = person.lastName;
+  document.getElementById("company-field").value = person.company;
+  document.getElementById("job-field").value = person.job;
+  document.getElementById("linkedin-field").value = person.linkedIn;
```
meagancheese: ' is preferred over " https://sites.google.com/a/google.com/dfa6/development/javascript-1/JavaScript-Style-Guide

src/main/webapp/profile.js
```diff
@@ -39,4 +61,4 @@ function makeEditable() {
   currentJob.setAttribute("hidden", true);
   const jobField = document.getElementById("job-field");
```
meagancheese: Same thing here, ' instead of ".

meagancheese (changes requested):
Small tweaks.

meagancheese (approved):
LGTM, will be curious what Alex says about no exception, not sure how the Optionals should be handled.

GrantComm (approved):
LGTM

src/main/webapp/logout.js
```diff
@@ -12,12 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// If user is logged in, returns a logout link. If not, redirects to home page.
 function supplyLogoutLink() {
   fetch('/login').then(response => response.json()).then(status => {
     if(status.loggedIn){
```
acmcarther: space between `if` and `(`.

src/main/webapp/profile.js
```diff
@@ -13,30 +13,52 @@
...
+  autofillForm(getUserIfRegistered());  
   prepareFormValidation();
 }
 
-// Allows certain fields in the profile to be edited, hides edit button, and displays update button. 
+// Returns a Person if they registered in the past. If not, redirect to  
+// registration page.
+function getUserIfRegistered(){
+  fetch('/login')
+  .then(response => response.json())
```
acmcarther: Indent this by two.

src/main/webapp/profile.js
```diff
@@ -13,30 +13,52 @@
...
+  .then(status => fetch(`/person?email=${status.email}`))
+  .then(response => {
+    if (response.redirected) {
```
acmcarther: Indent this by two.

src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -38,7 +39,7 @@ public void init(PersonDao personDao) {
     this.personDao = personDao;
   }
 
-  // Sends the request's contents to Datastore in the form of a new Person
+  // Sends the request's contents to Datastore in the form of a new Person.
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
acmcarther: What's the context that `post` is used?

I'm trying to discern whether you're planning to allow arbitrary people to `post` a new user. If so, that's suspicious.

src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -49,5 +50,19 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
             request.getParameter("company"),
             request.getParameter("job"),
             request.getParameter("linkedin")));
+    // TODO: redirect to home page and handle in JS.
+  }
+
+  // Returns the person the request's email belongs to.
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Optional<Person> personOpt = personDao.get(request.getParameter("email"));
```
acmcarther: This seems fishy. You're allowing any user to query for any other user by email, and sending them back the user's full name, company, job, and linked in address?

This data is technically PII (personally identifiable information), and probably should not just be returned to whomever asks for it.

It looks like you're intending to only have the user themselves fetch this. Can we more deeply integrate the authentication system to ensure that requester is *authenticated as* the user they're looking up?

src/main/webapp/profile.js
```diff
@@ -13,30 +13,52 @@
 // limitations under the License.
 
 function onProfileLoad() {
+  supplyLogoutLink();  
+  autofillForm(getUserIfRegistered());
```
acmcarther: I don't think this works the way you think it does.

getUserIfRegistered returns nothing here, but even if you were returning the promise, it would not be accessible as if it was a person. You need to chain a `.then()` off of it.

acmcarther (changes requested):
First pass

src/main/webapp/profile.js
apluscs: Why? The if is indented.

src/main/java/com/google/sps/servlets/PersonServlet.java
apluscs: lt's what happens after someone registers. Should l check if the request has all the necessary fields?

src/main/java/com/google/sps/servlets/PersonServlet.java
apluscs: l can add a check to ensure that. lt calls for some restructuring, however, so l'll put it in another PR (56).

src/main/webapp/profile.js
acmcarther: (because the `.then` needs to be indented more)

src/main/java/com/google/sps/servlets/PersonServlet.java
acmcarther: SGTM, thanks.

src/main/java/com/google/sps/servlets/PersonServlet.java
acmcarther: Hmm..

I'm not sure exactly what precautions make sense. Perhaps for now put a todo to "consider further restricting access to this API"?

src/main/webapp/profile.js
apluscs: Gotcha, done.

src/main/java/com/google/sps/servlets/PersonServlet.java
apluscs: Also depends on Person class, which needs to be refactored to Autovalue style. Another PR.

acmcarther:
I think you might have forgotten to re-request review, I'll re-review.

src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -49,5 +50,19 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
             request.getParameter("company"),
             request.getParameter("job"),
             request.getParameter("linkedin")));
+    // TODO: redirect to home page and handle in JS.
+  }
+
+  // Returns the person the request's email belongs to.
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
acmcarther, line 76: No tests?

acmcarther (changes requested):
Performing a redirect in getUserIfRegistered() seems poor form as it makes the function's return semantics really weird, but I doubt it actually matters in practice, so whatever.

I note that we don't seem to have any way to write tests for the javascript. Will discuss during standup today.

acmcarther:
What's the status of this PR? I think my last comment there ("No tests?") is the only comment. Is that proving to be a hassle?

apluscs:
Yes, l'm working on the tests. Could probably take a while.

src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -65,5 +67,26 @@ public void doPut(HttpServletRequest request, HttpServletResponse response) thro
...
+    // TODO: redirect to home page and handle in JS.
+  }
+
+  // Returns the person the request's email belongs to. If they aren't in Datastore, redirects to
+  // registration page. If the requestee is not the logged in user, throws a
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Ensure person logged in == person being requested.
+    String requesteeEmail = request.getParameter("email"),
+        userEmail = LogInServlet.getLoginInfo("/").email;
```
acmcarther: one declaration per statement.

acmcarther (approved):
One nit

PR #58, opened 2020-07-07
Simplify TimeRange.overlaps and TimeRange.contains implementations.

GrantComm:
Removed some redundant booleans in the implementation of those two functions, suggested by @apluscs.

### DIFF
--- a/src/main/java/com/google/sps/data/TimeRange.java
+++ b/src/main/java/com/google/sps/data/TimeRange.java
@@ -79,13 +79,14 @@ public final class TimeRange {
     //
     // Case 3: |---------|
     //            |---|
+    if (this.end == other.start || other.end == this.start) {
+      return false;
+    }
+
     if (this.start.isBefore(other.start)) {
       return this.contains(other.start);
     }
 
-    if (this.start.equals(other.start)) {
-      return this.end.isAfter(other.end);
-    }
     return other.contains(this.start);
   }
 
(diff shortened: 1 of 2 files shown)


acmcarther (approved):
I recommend a more specific PR title and description. 

When Jack and I look through PRs for final eval, we're going to have to look at this PR's code in order to tell what it's actually doing. A super clear title and description can help us a lot.

Here is how I might have worded this.

```
Simplify TimeRange.overlaps and contains implementations.

This PR removes some redundant booleans in the implementation of those two functions, suggested by @apluscs. Thanks!
```

src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -79,13 +79,14 @@ public boolean overlaps(TimeRange other) {
...
+    }
+
     if (this.start.isBefore(other.start)) {
       return this.contains(other.start);
     }
 
-    if (this.start.equals(other.start)) {
-      return this.end.isAfter(other.end);
-    }
     return other.contains(this.start);
```
apluscs: This function looks pretty good, but can be shortened if you think about what cases would mean no overlap.
There's only no overlap when other.end <= this.start OR other.start >= this.end. So in code:
``` ! (!other.end.isAfter(this.start) || !other.start.isBefore(this.end)) ``` = 
``` other.end.isAfter(this.start) && other.start.isBefore(this.end) ```

apluscs (changes requested):
One logic quibble.
Also, l remember you adding some tests. Are those going to be in another PR?

src/main/java/com/google/sps/data/TimeRange.java
GrantComm: @apluscs Yes, additional test will be in separate pr

apluscs:
Ok since you're adding more tests you should probably put them in the same PR as this one. Then we can know the changes pass these new tests (and also that the tests are more likely to be accurate).

apluscs (approved):
LGTM

PR #62, opened 2020-07-08
Quick format of things the autoformatter disagrees with

apluscs:
These things keep cluttering up my other PRs. Let's just merge them in.

### DIFF
--- a/src/main/java/com/google/sps/data/Availability.java
+++ b/src/main/java/com/google/sps/data/Availability.java
@@ -30,11 +30,7 @@ public abstract class Availability {
   public abstract LocalDate date();
 
   public static Availability create(String email, TimeRange when, LocalDate date) {
-    return builder()
-        .setEmail(email)
-        .setWhen(when)
-        .setDate(date)
-        .build();
+    return builder().setEmail(email).setWhen(when).setDate(date).build();
   }
 
   public static Builder builder() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/data/Availability.java
```diff
@@ -30,11 +30,7 @@
   public abstract LocalDate date();
 
   public static Availability create(String email, TimeRange when, LocalDate date) {
-    return builder()
-        .setEmail(email)
-        .setWhen(when)
-        .setDate(date)
-        .build();
+    return builder().setEmail(email).setWhen(when).setDate(date).build();
```
meagancheese, line 33: I think Alex preferred the vertically listed out version.

meagancheese (changes requested):

src/main/java/com/google/sps/data/Availability.java
apluscs: Yes but the formatter changes it back every time l compile it which is a nuisance because for every other PR l submit l have to manually change it back so it doesn't show up as a change. 
Alex wanted it vertically listed so that we can easily add more fields in the future. But when we do add more fields later the formatter will make it vertically listed if the line gets long enough.

meagancheese (approved):
LGTM

GrantComm (approved):
LGTM

acmcarther (approved):

PR #65, opened 2020-07-08
Connected profile page update to Person servlet

apluscs:


src/main/webapp/profile.html
```diff
@@ -45,28 +45,28 @@ <h1 class="text-center">Your Profile</h1>
...
             <!-- TODO: replace with user's actual email -->
-            <input type="text" readonly class="form-control-plaintext" id="user-email" value="tess@gmail.com">
+            <input type="text" readonly class="form-control-plaintext" id="user-email" value="tess@gmail.com" name="user-email">
           </div>
         </div>
         <div class="form-row">
           <div class="col-md-6 mb-3">
             <label for="first-name-field">First name</label>
-            <input type="text" readonly class="form-control-plaintext editable" id="first-name-field" placeholder="First name" value="Tess" required>
+            <input type="text" readonly class="form-control-plaintext editable" id="first-name-field" placeholder="First name" value="Tess" required name="first-name">
```
meagancheese: Long line

src/main/webapp/profile.html
```diff
@@ -45,28 +45,28 @@ <h1 class="text-center">Your Profile</h1>
...
           </div>
           <div class="col-md-6 mb-3">
             <label for="last-name-field">Last name</label>
-            <input type="text" readonly class="form-control-plaintext editable" id="last-name-field" placeholder="Last name" value="Ter" required>
+            <input type="text" readonly class="form-control-plaintext editable" id="last-name-field" placeholder="Last name" value="Ter" required name="last-name">
```
meagancheese: Long line

src/main/webapp/profile.html
```diff
@@ -82,13 +82,12 @@ <h1 class="text-center">Your Profile</h1>
         <div class="form-row">
           <div class="col-md-12 mb-3">
             <label for="linkedin-field">LinkedIn URL</label>
-            <input type="text" readonly class="form-control-plaintext editable" id="linkedin-field" placeholder="https://www.linkedin.com/" value="">
+            <input type="text" readonly class="form-control-plaintext editable" id="linkedin-field" placeholder="https://www.linkedin.com/" value="" name="linkedin">
```
meagancheese: Long line

src/main/webapp/profile.html
```diff
@@ -82,13 +82,12 @@ <h1 class="text-center">Your Profile</h1>
...
           </div>
         </div>
         <div class="form-group row">
           <div class="col-sm-10">
-            <!-- TODO: Have this update button actually update profile info to Datastore. -->            
-            <button type="submit" id="update-button" hidden class="btn btn-primary">Update</button>
+            <button class="btn btn-primary" id="update-button" type="submit" hidden>Update Profile</button>
```
meagancheese: There's no function being called here which makes me suspicious of the PR title.

meagancheese (changes requested):
Some formatting and then a note on PR title.

src/main/webapp/profile.html
apluscs: lt's a form that submits to a servlet. 
Edit: l see that l need to call a JS function to redirect to the homepage instead of leading them to the /register blank page after they update. Will do.

src/main/webapp/profile.js
```diff
@@ -14,14 +14,19 @@
 
 function onProfileLoad() {
   prepareFormValidation();
```
meagancheese, line 20: Where is this method and why is it before supplyLogoutLink?

src/main/webapp/profile.js
```diff
@@ -14,14 +14,19 @@
...
-
-function onProfileLoad() {
-  supplyLogoutLink();
+// Submits profile form to Datastore.
+function submitProfileForm(methodType, redirectUrl) {
+  fetch(`/person?user-email=${$("#user-email").val()}&first-name=${$("#first-name-field").val()}`+
+    `&last-name=${$("#last-name-field").val()}&company=${$("#company-field").val()}&job=`+
+    `${$("#job-field").val()}&linkedin=${$("#linkedin-field").val()}`, 
+    {method: methodType});
+  window.location.replace(redirectUrl);
```
meagancheese: This needs to only execute after the fetch goes through, so either add an await or .then promise capture.

meagancheese (changes requested):

src/main/webapp/profile.js
apluscs: lts in register.js

meagancheese (approved):
LGTM

GrantComm (approved):

src/main/webapp/profile.js
```diff
@@ -13,15 +13,21 @@
...
-
-function onProfileLoad() {
-  supplyLogoutLink();
+// Submits profile form to Datastore.
+function submitProfileForm(methodType, redirectUrl) {
+  fetch(`/person?user-email=${$("#user-email").val()}&first-name=${$("#first-name-field").val()}`+
+    `&last-name=${$("#last-name-field").val()}&company=${$("#company-field").val()}&job=`+
+    `${$("#job-field").val()}&linkedin=${$("#linkedin-field").val()}`, 
+    {method: methodType})
+    .then(window.location.replace(redirectUrl));
```
jalexanderqed: Is there a strong reason you prefer having this URL-encoded? If not, it should be JSON text in the request body (see Meagan's availability update code for an example). The "user" JSON object should have a Java equivalent class.

jalexanderqed: It probably make sense to explicitly handle any errors here and alert to the user if this fails.

jalexanderqed (changes requested):

src/main/webapp/profile.js
apluscs: Changed everything to use the request body strategy.

src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -38,36 +43,38 @@ public void init() {
...
-            request.getParameter("last-name"),
-            request.getParameter("company"),
-            request.getParameter("job"),
-            request.getParameter("linkedin")));
+            personJson.get("userEmail").getAsString(),
+            personJson.get("firstName").getAsString(),
+            personJson.get("lastName").getAsString(),
+            personJson.get("company").getAsString(),
+            personJson.get("job").getAsString(),
+            personJson.get("linkedin").getAsString()));
```
jalexanderqed: Please change this as discussed in the meeting (create a PersonJson or similar name class).

src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -38,36 +43,38 @@ public void init() {
...
-    personDao.update(
-        Person.create(
-            request.getParameter("user-email"),
-            request.getParameter("first-name"),
-            request.getParameter("last-name"),
-            request.getParameter("company"),
-            request.getParameter("job"),
-            request.getParameter("linkedin")));
-    // TODO: redirect to home page and handle in JS.
+    doPost(request, response);
```
jalexanderqed: It seems like this should call personDao.update(), not create it every time?

jalexanderqed (changes requested):

src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -38,36 +40,46 @@ public void init() {
...
-    // TODO: redirect to home page and handle in JS.
+    try {
+      PersonRequest personRequest =
+          new Gson().fromJson(getJsonString(request), PersonRequest.class);
+      personDao.update(Person.create(personRequest));
+    } catch (Exception JsonSyntaxException) {
+      response.sendError(400);
+      return;
+    }
+  }
```
jalexanderqed: All of these need to check authentication. As it is, any user could construct a request to modify any other user.

jalexanderqed (changes requested):

jalexanderqed (approved):

PR #66, opened 2020-07-08
Added Availability Dao lnterface

apluscs:


GrantComm (approved):
LGTM

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.List;
+
+/**
+ * AvailabilityDao includes the basic methods anything managing Interviewer Availability entities
+ * must support.
+ */
+public interface AvailabilityDao {
+  // Returns a list of all Availability's ranging from minTime to maxTime of a user.
+  public List<Availability> getInRangeForUser(String email, String minTime, String maxTime);
```
meagancheese: If minTime and maxTime are UTC Strings, explicitly state that.

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -0,0 +1,35 @@
...
+
+  // Returns all Availability's across all users ranging from minTime to maxTime.
+  public List<Availability> getInRangeForAll(String minTime, String maxTime);
```
meagancheese: Same here.

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -0,0 +1,35 @@
...
+
+  // Puts an Availability object into storage.
+  public void put(Availability availability);
+
+  // Deletes all Availability's entities for a user ranging from minTime to maxTime.
```
meagancheese: Availability entities

meagancheese: Also, I thought with our current implementation this wouldn't depend on a min and max time but rather deletes all Availability entities.

meagancheese (changes requested):

src/main/java/com/google/sps/data/AvailabilityDao.java
apluscs: l feel adding more than one week is a high priority, something that's likely to get done. l want to prepare for that as much as possible.

meagancheese (approved):
LGTM

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -0,0 +1,38 @@
...
+  // Returns a list of all Availability's ranging from minTime to maxTime of a user.
+  // minTime and maxTime are UTC encodings.
+  public List<Availability> getInRangeForUser(String email, String minTime, String maxTime);
+
+  // Returns all Availability's across all users ranging from minTime to maxTime.
+  // minTime and maxTime are UTC encodings.  
+  public List<Availability> getInRangeForAll(String minTime, String maxTime);
+
+  // Puts an Availability object into storage.
+  public void put(Availability availability);
```
jalexanderqed: Please switch to create/update

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+
+/**
+ * AvailabilityDao includes the basic methods anything managing Interviewer Availability entities
+ * must support.
+ */
+public interface AvailabilityDao {
+  // Returns a list of all Availability's ranging from minTime to maxTime of a user.
+  // minTime and maxTime are UTC encodings.
+  public List<Availability> getInRangeForUser(String email, String minTime, String maxTime);
```
jalexanderqed: Like in ScheduledInterviewDao, there should be a `get(long id)` method for getting an availability object by its Datastore-created ID. You may not need it, but if these DAO objects ever inherit from a parent interface, they will all need a `get()` method.

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/AvailabilityDao.java
apluscs: Since we're doing the delete and replace all strategy, we have no use for update. l will add create through.

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -0,0 +1,44 @@
...
+  public List<Availability> getInRangeForAll(String minTime, String maxTime);
+
+  // Returns the Availability entity with specified id.
+  public Optional<Availability> get(long id);
+
+  // Adds an Availability object into storage.
+  public void create(Availability availability);
+
+  // Updates the specified id with the new availability.
+  public Optional<Availability> update(long id, Availability availability);
```
jalexanderqed: `Availability` should have an ID field so that you are not required to pass this `id` parameter. That can be done in a separate PR since there is no implementation here. Just remove the `id` parameter.

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -0,0 +1,44 @@
...
+  public List<Availability> getInRangeForAll(long minTime, long maxTime);
+
+  // Returns the Availability entity with specified id.
+  public Optional<Availability> get(long id);
+
+  // Adds an Availability object into storage.
+  public void create(Availability availability);
+
+  // Updates the specified id with the new availability.
+  public Optional<Availability> update(Availability availability);
```
jalexanderqed: Why does this return something?

jalexanderqed (approved):

src/main/java/com/google/sps/data/AvailabilityDao.java
apluscs: lt shouldn't oops. l'll change it to void.

PR #71, opened 2020-07-09
Refactored PersonDao to use the create/update model.

apluscs:
This required changes in DatastorePersonDao and PersonServlet as well.

### DIFF
--- a/src/main/java/com/google/sps/data/DatastorePersonDao.java
+++ b/src/main/java/com/google/sps/data/DatastorePersonDao.java
@@ -44,7 +44,15 @@ public class DatastorePersonDao implements PersonDao {
    * We make an entity in Datastore with person's fields as properties and their email as the key.
    */
   @Override
-  public void put(Person person) {
+  public void create(Person person) {
+    datastore.put(personToEntity(person));
+  }
+
+  /**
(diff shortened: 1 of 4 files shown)


GrantComm (approved):
LGTM

src/main/java/com/google/sps/data/DatastorePersonDao.java
```diff
@@ -44,7 +44,15 @@ public DatastorePersonDao() {
...
-  public void put(Person person) {
+  public void create(Person person) {
+    datastore.put(personToEntity(person));
+  }
+
+  /**
+   * We update an entity in Datastore with person's fields as properties and their email as the key.
+   */
+  @Override
+  public void update(Person person) {
```
meagancheese, line 55: How is this updating? Does it replace the old one because they both have the same keys (email) ?

meagancheese (approved):
One q, but otherwise LGTM.

acmcarther (approved):

src/main/java/com/google/sps/data/DatastorePersonDao.java
apluscs: Yes, exactly.

PR #77, opened 2020-07-10
Creates a servlet that stores a scheduledInterview in Datastore

GrantComm:
Tested using hard-coded request in JavaScript (changes to JS files will go in separate pr)


src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
```diff
@@ -0,0 +1,63 @@
...
+
+package com.google.sps.servlets;
+
+import java.io.IOException;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
```
apluscs, line 27: Need to remove these blank lines.
https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
```diff
@@ -0,0 +1,63 @@
...
+    this.scheduledInterviewDao = scheduledInterviewDao;
+  }
+
+  // Sends the request's contents to Datastore in the form of a new ScheduledInterview.
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/MM/yyyy");
+    ScheduledInterview scheduledInterview =
+        ScheduledInterview.create(
+            0,
```
apluscs: Nit: l think Jack suggested -1, because 0 could still be a legit id in Datastore.

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
```diff
@@ -0,0 +1,63 @@
...
+
+  // TODO: add a FakeScheduledInterviewDao class so this will become useful
+  public void init(ScheduledInterviewDao scheduledInterviewDao) {
+    this.scheduledInterviewDao = scheduledInterviewDao;
+  }
+
+  // Sends the request's contents to Datastore in the form of a new ScheduledInterview.
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/MM/yyyy");
```
apluscs: When is this ever used?

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
```diff
@@ -0,0 +1,63 @@
...
+    ScheduledInterview scheduledInterview =
+        ScheduledInterview.create(
+            0,
+            new TimeRange(
+                Instant.parse(request.getParameter("startTime")),
+                Instant.parse(request.getParameter("endTime"))),
+            LocalDate.parse(request.getParameter("date")),
```
apluscs: Don't need date in Datastore, but that should be in another PR.

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  // Sends the request's contents to Datastore in the form of a new ScheduledInterview.
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/MM/yyyy");
+    ScheduledInterview scheduledInterview =
+        ScheduledInterview.create(
+            0,
+            new TimeRange(
```
apluscs, line 49: Datastore can't represent TimeRange. What l'm doing with Availability is converting TimeRange's lnstants to longs using https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#toEpochMilli--

and storing longs in Datastore. 
l'm thinking on the front end we should have these longs stored in some data- attribute in the HTML and we just pass them in as longs here.
lt might honestly be more worth it to get some of the front end set up first, so we know what's being passed in exactly.

apluscs (changes requested):

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
GrantComm: I'm currently storing the instants as strings in Datastore via the create method in DatastoreScheduledInterview.java but I can change them to longs instead

apluscs: Hm yeah it'd be better if it matched with Availability entities

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.sps.data.ScheduledInterviewDao;
+import com.google.sps.data.TimeRange;
+import java.io.IOException;
+import java.time.Instant;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/new-scheduled-interview")
```
apluscs, line 28: This is going to be used eventually to get a user's scheduled interviews as well right? So remove the `new-`

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
```diff
@@ -0,0 +1,56 @@
...
+public class NewScheduledInterviewServlet extends HttpServlet {
```
apluscs, line 29: New here is questionable too

apluscs (changes requested):

src/main/java/com/google/sps/servlets/NewScheduledInterviewServlet.java
GrantComm: I have them as two separate files, there's another servlet for listing scheduledInterviews

PR #81, opened 2020-07-10
Updating Availability 1: Creates the JS onclick method that collects the selected time slots and forms the Request to the backend.

meagancheese:
In an effort to have smaller PRs, here is the javascript and jsp file changes that execute the preparation of data to send in a fetch request.

### DIFF
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -40,5 +40,5 @@ function availabilityTableDiv() {
 }
 
 function updateAvailability() {
-  
+  let selected = document.getElementsByClassName('table-success');
 }
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -38,3 +38,7 @@ function browserTimezoneOffset() {
 function availabilityTableDiv() {
   return document.getElementById('table-container');
 }
+
+function updateAvailability() {
+  
+}
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -41,9 +41,13 @@ function availabilityTableDiv() {
 
 function updateAvailability() {
   let selectedSlots = document.getElementsByClassName('table-success');
+  let firstSlot = document.getElementsByTagName('tbody').item(0)
+    .firstElementChild.firstElementChild.getAttribute('data-utc');
+  let lastSlot = document.getElementsByTagName('tbody').item(0)
+    .lastElementChild.lastElementChild.getAttribute('data-utc');
   let requestBody = [
-    document.getElementById('first-slot').getAttribute('data-utc'),
-    document.getElementById('last-slot').getAttribute('data-utc')
+    firstSlot,
+    lastSlot
   ].concat(Array.from(selectedSlots).map(s => s.getAttribute('data-utc')));
   console.log(requestBody);
   let request = new Request('/availability', {method:'PUT', body:requestBody});
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -51,4 +51,6 @@ function updateAvailability() {
     requestBody.push(utc);
   }
   console.log(requestBody);
+  let request = new Request('/availability', {method:'PUT', body:requestBody});
+  
 }
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -42,6 +42,10 @@ function availabilityTableDiv() {
 function updateAvailability() {
   let selectedSlots = document.getElementsByClassName('table-success');
   let requestBody;
+  let firstSlot = {firstSlot : document.getElementById('first-slot').getAttrubute('data-utc')};
+  let lastSlot = {lastSlot : document.getElementById('last-slot').getAttribute('data-utc')};
+  requestBody.put(firstSlot);
+  requestBody.put(lastSlot);
   for (let slot of selectedSlots) {
     utc = slot.getAttribute('data-utc');
     requestBody.put(utc);
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -50,5 +50,5 @@ function updateAvailability() {
     lastSlot
   ].concat(Array.from(selectedSlots).map(s => s.getAttribute('data-utc')));
   
-  let request = new Request('/availability', {method:'PUT', body:JSON.stringify(utcTimestamps)});
+  let request = new Request('/availability', {method: 'PUT', body: JSON.stringify(utcTimestamps)});
 }
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -41,13 +41,14 @@ function availabilityTableDiv() {
 
 function updateAvailability() {
   let selectedSlots = document.getElementsByClassName('table-success');
-  let requestBody;
-  let firstSlot = {firstSlot : document.getElementById('first-slot').getAttrubute('data-utc')};
+  let requestBody = [];
+  let firstSlot = {firstSlot : document.getElementById('first-slot').getAttribute('data-utc')};
   let lastSlot = {lastSlot : document.getElementById('last-slot').getAttribute('data-utc')};
-  requestBody.put(firstSlot);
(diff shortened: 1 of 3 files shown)


src/main/webapp/availabilityTable.jsp
```diff
@@ -21,7 +21,17 @@
   <tbody>
```
apluscs: lnstead of writing out all the specifics to navigate around the first-slot last-slot situation, can we make a simple for loop? And add data- attributes to the table-success table that have the minTime and maxTime?

apluscs (approved):
We can simplify this

apluscs (changes requested):
Accidentally hit approve.

src/main/webapp/availabilityTable.jsp
meagancheese: It won't really be element specific then though. @acmcarther  @jalexanderqed what do you guys think? Prioritize making a nicer jsp page or only put ids on two slots?

src/main/webapp/availability.js
```diff
@@ -38,3 +38,22 @@ function browserTimezoneOffset() {
 function availabilityTableDiv() {
   return document.getElementById('table-container');
 }
+
+function updateAvailability() {
+  let selectedSlots = document.getElementsByClassName('table-success');
+  let requestBody = [];
+  let firstAndLastSlots = [];
```
acmcarther: Nit: You can remove basically all of these variable declarations by using a declarative style.

```javascript
let requestBody = [
  [
    document.getElementById('first-slot').getAttribute('data-utc'),
    document.getElementById('last-slot').getAttribute('data-utc'),
  ],
  selectedSlots.map(s => s.getAttribute('data-utc'))
]
```

After writing that though, i suspect that's not the logic you actually had in mind. Maybe this instead?

```javascript
let request_body = [
  document.getElementById('first-slot').getAttribute('data-utc'),
  document.getElementById('last-slot').getAttribute('data-utc'),
].concat(selectedSlots.map(s => s.getAttribute('data-utc'));
```

My preferred way to implement this would be to collect all the DOM elements and then get the attributes though

```javascript
let request_body = selectedSlots
  .concat([
     document.getElementById('first-slot'),
     document.getElementById('last-slot'),
  ])
  .map(el => el.getAttribute('data-utc'));

File: src/main/webapp/availabilityTable.jsp [hunk 3]
```diff
@@ -21,14 +21,42 @@
...
-    <c:forEach var = "i" begin = "0" end = "${pageScope.list.get(0).size() - 1}">
+    <tr>
+      <td id="first-slot" onclick="toggleTile(this)" 
+        data-utc="${list.get(0).get(0).utcEncoding()}" 
+        class="${list.get(0).get(0).selected() ? 'table-success' : ''}">
+        ${list.get(0).get(0).time()}
+      </td>
+      <c:forEach var = "i" begin = "1" end = "6">
+        <td onclick="toggleTile(this)" data-utc="${list.get(i).get(0).utcEncoding()}" 
+          class="${list.get(i).get(0).selected() ? 'table-success' : ''}">
```
acmcarther: I'm going to break my "dont complain about something without a suggestion rule" this time.

This, and especially the calls below, are a mess. Please refactor this.

acmcarther (changes requested):

src/main/webapp/availability.js
meagancheese: Just a note: .map does not work on HTMLCollections, used Array.from() to solve that problem.

src/main/webapp/availabilityTable.jsp
meagancheese: Fixed it with children accessors.

src/main/webapp/availabilityTable.jsp
meagancheese: Fixed.

meagancheese:
This is ready for re-review.

apluscs (approved):

GrantComm (approved):

src/main/webapp/availability.js
```diff
@@ -38,3 +38,17 @@ function browserTimezoneOffset() {
...
+  let firstSlot = document.getElementsByTagName('tbody').item(0)
+    .firstElementChild.firstElementChild.getAttribute('data-utc');
+  let lastSlot = document.getElementsByTagName('tbody').item(0)
+    .lastElementChild.lastElementChild.getAttribute('data-utc');
+  let utcTimestamps = [
+    firstSlot,
+    lastSlot
+  ].concat(Array.from(selectedSlots).map(s => s.getAttribute('data-utc')));
+  
+  let request = new Request('/availability', {method:'PUT', body:JSON.stringify(utcTimestamps)});
```
jalexanderqed: I think there should be spaces after each of the colons here.

src/main/webapp/availability.js
```diff
@@ -38,3 +38,17 @@ function browserTimezoneOffset() {
...
   return document.getElementById('table-container');
 }
+
+function updateAvailability() {
+  let selectedSlots = document.getElementsByClassName('table-success');
+  let firstSlot = document.getElementsByTagName('tbody').item(0)
+    .firstElementChild.firstElementChild.getAttribute('data-utc');
+  let lastSlot = document.getElementsByTagName('tbody').item(0)
+    .lastElementChild.lastElementChild.getAttribute('data-utc');
+  let utcTimestamps = [
```
jalexanderqed: This system of having the first two elements of the array be the period boundaries is a bit strange and fragile. It will be even more surprising to people reading the Java code. It would probably be better to have an object (which will be mirrored in Java for use with Gson) that has the fields `firstSlot`, `lastSlot`, and `selectedSlots`. You would construct it as:

```
let requestObj = {
  firstSlot: firstSlot,
  lastSlot: lastSlot,
  selectedSlots: selectedSlots,
}
```

src/main/webapp/availability.js
```diff
@@ -38,3 +38,17 @@ function browserTimezoneOffset() {
 function availabilityTableDiv() {
   return document.getElementById('table-container');
 }
+
+function updateAvailability() {
+  let selectedSlots = document.getElementsByClassName('table-success');
```
jalexanderqed: Getting the elements by a class used for formatting is fragile. If anyone ever added a new table that used this class to this page, this code would break. You should probably add an additional class of your own devising (like `selected-time-slot`) to each selected element in addition to `table-success`.

jalexanderqed (changes requested):
Please change the name of the PR to something more descriptive. It will show up when looking at the history of a file, and it's helpful if it clearly describes the change.

src/main/webapp/availability.js
meagancheese: Done

src/main/webapp/availability.js
meagancheese: Done

src/main/webapp/availability.js
meagancheese: Done

meagancheese:
This is ready for re-review.

src/main/webapp/availability.js
```diff
@@ -17,9 +17,14 @@ function onAvailabilityLoad() {
   loadAvailabilityTable(availabilityTableDiv(), browserTimezoneOffset());
 }
 
-// Toggles a tile from green to white and vice versa when clicked.
+// Toggles a tile from selected (green) to un-selected (white) and vice versa when clicked.
 function toggleTile(tile) {
-  tile.classList.toggle('table-success');
+  let classList = tile.classList;
+  if (classList.contains('table-success') && classList.contains('selected-time-slot')) {
```
acmcarther, line 23: Can we change `table-success` in this PR to something more descriptive as well?

acmcarther (approved):
Generally LGTM, just one nit

src/main/webapp/availability.js
meagancheese: table-success is the Bootstrap class for a green table entry.

jalexanderqed (approved):

PR #86, opened 2020-07-10
Updating Availability 2: Creates an Availability Servlet and parses the HTTP request in doPut.

meagancheese:
Sends the fetch request and parses the request body into a PutAvailabilityRequest object.


src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/availability")
+public class AvailabilityServlet extends HttpServlet {
+  @Override
+  public void doPut(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    BufferedReader reader = request.getReader();
```
apluscs, line 32: Yikes! Maybe this will help with deserializing JSON: https://www.java67.com/2016/10/3-ways-to-convert-string-to-json-object-in-java.html

src/main/webapp/availability.js
```diff
@@ -49,6 +49,6 @@ function updateAvailability() {
     firstSlot,
     lastSlot
   ].concat(Array.from(selectedSlots).map(s => s.getAttribute('data-utc')));
-  
   let request = new Request('/availability', {method:'PUT', body:JSON.stringify(utcTimestamps)});
+  fetch(request).then(unused => {loadAvailabilityTable(availabilityTableDiv(), browserTimezoneOffset())});
```
apluscs, line 60: Question: what is this doing? Updating the table with availabilities?
lf so we might experience some inconsistencies since we're not updating by key or anything, so eventual consistency might cause it show inaccurate results. :/

apluscs (changes requested):

src/main/webapp/availability.js
meagancheese: This is reloading the table, which should be done after an update.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
meagancheese: This is necessary because we're going from a HTTPServletRequest to an array of Strings. There is no readily available JSON String, thus the need for this process.

src/main/webapp/availability.js
meagancheese: It's within a promise, which should help somewhat. I'm not sure how much we should be compensating for datastore latency.

apluscs (approved):

GrantComm (approved):

meagancheese:
This needed a re-vamp to include the new JSONConverter class, please re-review.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -0,0 +1,41 @@
...
+    StringBuffer buffer = new StringBuffer();
+    String timestamps = null;
+
+    while ((timestamps = reader.readLine()) != null) buffer.append(timestamps);
+    String jsonString = buffer.toString();
```
apluscs: l used `response.getContentAsString();` to get the JSON string of the response. Would this help?

apluscs (changes requested):
AvailabilityJSONConverter looks good, but doPut() can be simplified l think.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
meagancheese: Wrong context (Claire is using this in mock testing).

apluscs (approved):

GrantComm (approved):

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/availability")
+public class AvailabilityServlet extends HttpServlet {
+  @Override
+  public void doPut(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    BufferedReader reader = request.getReader();
+    StringBuffer buffer = new StringBuffer();
+    String timestamps = null;
```
acmcarther: Timestamps here is misleading. This is actually "payloadLine" (or something approximately like that), which could be organized in any JSON syntactically valid fashion provided it results in an object with fields matching `AvailabilityJSONConverter`.

src/main/java/com/google/sps/data/AvailabilityJSONConverter.java
```diff
@@ -0,0 +1,42 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.ArrayList;
+
+public class AvailabilityJSONConverter {
```
acmcarther: Nit: This class doesn't convert anything. I understand where the name came from, but as a rule I'd suggest sticking close to the most literal definition for what it is. By my reckoning, it's a `PutAvailabilityRequest`.

Etymology:

Put-Availability-Request
- Put: It's a `Put` request, and it will probably not be reused outside of the `Put` context. You're not going to `Get` using this.
- Availability: (Self explanatory)
- Request: Indicates to reader that this is directly coupled to REST HTTP api, and probably received by some servlet. You're probably not going to send something exactly like this back (as in, a Response).

acmcarther (changes requested):
Nits, mostly lgtm

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
meagancheese: Done

src/main/java/com/google/sps/data/AvailabilityJSONConverter.java
meagancheese: Done

meagancheese:
Ready for re-review.

acmcarther (changes requested):
Discussed in chat:

- Update doPut to return a 400 if request json is not parseable.
- Add tests for valid and invalid requests.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      utcEncodings = new Gson().fromJson(jsonString, PutAvailabilityRequest.class);
+    } catch (Exception JsonSyntaxException) {
+      response.sendError(400);
+      return;
+    }
+    if (!utcEncodings.allFieldsPopulated()) {
+      response.sendError(400);
+      return;
+    }
```
acmcarther, line 49: Add TODO for submitting the availabilities into DataStore.

src/test/java/com/google/sps/data/AvailabilityServletTest.java
```diff
@@ -0,0 +1,80 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import static org.junit.Assert.*;
```
acmcarther: `*` imports are discouraged, import the things you need directly.

I don't see where these are being used anyway, so you might be able to just drop this..

acmcarther (approved):
Two nits

src/test/java/com/google/sps/data/AvailabilityServletTest.java
meagancheese: Done

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
meagancheese: Done

PR #95, opened 2020-07-14
Implements the Availability Datastore DAO and tests it thoroughly.

meagancheese:


src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -0,0 +1,174 @@
...
+    availabilityEntity.setProperty("scheduled", avail.scheduled());
+    return availabilityEntity;
+  }
+
+  static Entity availabilityToUpdatedEntity(Availability avail) {
+    Entity availabilityEntity = new Entity("Availability", avail.id());
+    availabilityEntity.setProperty("email", avail.email());
+    availabilityEntity.setProperty("startTime", avail.when().start().toEpochMilli());
+    availabilityEntity.setProperty("endTime", avail.when().end().toEpochMilli());
+    availabilityEntity.setProperty("scheduled", avail.scheduled());
```
apluscs, line 93: Duplication. Probably best to abstract this out to another function.

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -0,0 +1,174 @@
...
+    List<Availability> availability = new ArrayList<Availability>();
+    for (Entity entity : entities) {
+      availability.add(entityToAvailability(entity));
+    }
+    return availability;
+  }
+
+  // Returns a list of Scheduled-over Availabilities ranging from minTime to maxTime of a user.
+  // minTime and maxTime are in milliseconds.
+  public List<Availability> getScheduledInRangeForUser(String email, long minTime, long maxTime) {
```
apluscs, line 141: Question: When will this be used?

src/test/java/com/google/sps/data/DatastoreAvailabilityDaoTest.java
```diff
@@ -0,0 +1,369 @@
...
+            availabilities.get(1).id(),
+            false);
+    List<Availability> expectedAvailabilities = new ArrayList<Availability>();
+    expectedAvailabilities.add(expectedAvailabilityOne);
+    expectedAvailabilities.add(expectedAvailabilityTwo);
+    Assert.assertEquals(expectedAvailabilities, actual);
+  }
+
+  @Test
+  public void getsScheduledAvailabilitiesForUserInRange() {
```
apluscs, line 281: Takes a lot of mental effort to figure out what these tests are doing, especially since their names are pretty similar. Could you put comments above?

apluscs (changes requested):

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: For determining whether or not an availability time slot is scheduled while loading the availability page.

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: Done

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
apluscs: We can just look at the scheduled flag for the user's availability right?

apluscs (approved):
Qualms about `getScheduledInRangeForUser` but other than LGTM.

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  static Entity availabilityToUpdatedEntity(Availability avail) {
+    Entity availabilityEntity = new Entity("Availability", avail.id());
+    setEntityProperties(availabilityEntity, avail);
+    return availabilityEntity;
+  }
+
+  static void setEntityProperties(Entity availabilityEntity, Availability avail) {
+    availabilityEntity.setProperty("email", avail.email());
```
apluscs: After looking at another PR, l realize this function goes against something Alex warned me against. That is, a function that modifies a param instead of returning it. Here it might be fine because it's more explicit in the function name but not sure. @acmcarther  @jalexanderqed

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: That's what this does though, it's just a matter of putting as much access and filter logic in the DAO as we can.

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -0,0 +1,176 @@
...
+        TimeRange.fromStartEnd(
+            Instant.ofEpochMilli((long) availabilityEntity.getProperty("startTime")),
+            Instant.ofEpochMilli((long) availabilityEntity.getProperty("endTime"))),
+        availabilityEntity.getKey().getId(),
+        (boolean) availabilityEntity.getProperty("scheduled"));
+  }
+
+  // Deletes all Availability entities for a user ranging from minTime to maxTime.
+  // minTime and maxTime are in milliseconds.
+  public void deleteInRangeForUser(String email, long minTime, long maxTime) {
```
GrantComm, line 110: Shouldn't this method include an @Override annotation ?

GrantComm (changes requested):
Add @Override annotations for all methods that you implemented from the interface

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: Done

GrantComm (approved):

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
acmcarther: Update this function to create and return an Entity from the input Availability

acmcarther: Or just combine availabilityToNewEntity and availabilityToUpdatedEntity.

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -0,0 +1,180 @@
...
+    Filter userFilter = new FilterPredicate("email", FilterOperator.EQUAL, email);
+    List<Entity> entities = getEntitiesInRange(minTime, maxTime, Optional.of(userFilter));
+    List<Key> keyList = new ArrayList<>();
+    for (Entity entity : entities) {
+      keyList.add(entity.getKey());
+    }
+    for (Key key : keyList) {
+      Transaction txn = datastore.beginTransaction();
+      datastore.delete(txn, key);
```
acmcarther, line 120: You're doing this this way to avoid cross-group transactions, right? Maybe add a comment here?

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -0,0 +1,180 @@
...
+    List<Availability> availability = new ArrayList<Availability>();
+    for (Entity entity : entities) {
+      availability.add(entityToAvailability(entity));
+    }
+    return availability;
+  }
+
+  private List<Entity> getEntitiesInRange(long minTime, long maxTime, Optional<Filter> filterOpt) {
+    Filter
+        minFilter = new FilterPredicate("startTime", FilterOperator.GREATER_THAN_OR_EQUAL, minTime),
```
acmcarther: One variable declaration per statement.

acmcarther (approved):
Couple of nits

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: The Entity constructors used in these are different, so I guess I'll just put this code back within the two methods.

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: Added

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: Done

PR #97, opened 2020-07-14
Changes DatastoreScheduledInterview update() method to use datastore put() and tests its functionality

GrantComm:
Modified the DataStoreScheduledInterviewDao.java file to include a method that changes an entity's fields and added tests for each method in the file. Formatting change in DatastoreAvailabilityDao.java is in pr #100

### DIFF
--- a/src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
+++ b/src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
@@ -91,7 +91,7 @@ public class DatastoreScheduledInterviewDao implements ScheduledInterviewDao {
   /** Updates an entity in datastore. */
   @Override
   public void update(ScheduledInterview scheduledInterview) {
-    datastore.put(scheduledInterviewToUpdateEntity(scheduledInterview));
+    datastore.put(scheduledInterviewToEntityForUpdate(scheduledInterview));
   }
 
   /** Deletes an entity in datastore. */
@@ -124,7 +124,7 @@ public class DatastoreScheduledInterviewDao implements ScheduledInterviewDao {
   }
 
   /** Creates a scheduledInterview Entity with the updated fields and id of a scheduledInterview */
-  public Entity scheduledInterviewToUpdateEntity(ScheduledInterview scheduledInterview) {
+  public Entity scheduledInterviewToEntityForUpdate(ScheduledInterview scheduledInterview) {
     Entity scheduledInterviewEntity = new Entity("ScheduledInterview", scheduledInterview.id());
     scheduledInterviewEntity.setProperty(
         "startTime", scheduledInterview.when().start().toEpochMilli());
--- a/src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
+++ b/src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
@@ -91,8 +91,7 @@ public class DatastoreScheduledInterviewDao implements ScheduledInterviewDao {
   /** Updates an entity in datastore. */
   @Override
   public void update(ScheduledInterview scheduledInterview) {
-    delete(scheduledInterview.id());
-    create(scheduledInterview);
+    datastore.put(scheduledInterviewToUpdateEntity(scheduledInterview));
   }
 
   /** Deletes an entity in datastore. */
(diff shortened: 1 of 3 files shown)


meagancheese (approved):
LGTM

src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
```diff
@@ -116,11 +115,24 @@ public ScheduledInterview entityToScheduledInterview(Entity scheduledInterviewEn
...
+  /** Updates a scheduledInterview Entity from a scheduledInterview object. */
+  public Entity scheduledInterviewToUpdateEntity(ScheduledInterview scheduledInterview) {
+    Entity scheduledInterviewEntity = new Entity("ScheduledInterview", scheduledInterview.id());
+    setEntityProperties(scheduledInterviewEntity, scheduledInterview);
+    return scheduledInterviewEntity;
+  }
+
+  /** Sets the properties of a scheduledInterview Entity. */
+  static void setEntityProperties(
+      Entity scheduledInterviewEntity, ScheduledInterview scheduledInterview) {
```
apluscs: This function goes against something Alex warned me against. That is, a function that modifies a param instead of returning it. Here it might be fine because it's more explicit in the function name but not sure. We can come back after we get some more info.

apluscs (approved):
One nit, but LGTM.

src/test/java/com/google/sps/data/DatastoreScheduledInterviewTest.java
```diff
@@ -0,0 +1,180 @@
...
+            "user3@mail.com");
+    tester.update(updatedStoredScheduledInterview);
+    Entity updatedEntity = datastore.prepare(new Query("ScheduledInterview")).asSingleEntity();
+    ScheduledInterview updatedScheduledInterview = tester.entityToScheduledInterview(updatedEntity);
+    Assert.assertEquals(updatedStoredScheduledInterview, updatedScheduledInterview);
+  }
+
+  // Tests retrieving a scheduledInterview from Datastore.
+  @Test
+  public void getsScheduledInterview() {
```
meagancheese, line 163: You should add a test where it is not within datastore to make sure an empty Optional is returned.

meagancheese (changes requested):
One more test, otherwise looks good.

src/test/java/com/google/sps/data/DatastoreScheduledInterviewTest.java
GrantComm: Done

src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
meagancheese: So the conclusion is to put this code back within the two methods because creation of the Entity, property setting, and returning should all be within the same method.

meagancheese (approved):
Small fix (delete setProperties), otherwise LGTM.

src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
GrantComm: Done

src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
```diff
@@ -123,4 +122,15 @@ public Entity scheduledInterviewToEntity(ScheduledInterview scheduledInterview)
     scheduledInterviewEntity.setProperty("interviewee", scheduledInterview.intervieweeEmail());
     return scheduledInterviewEntity;
   }
+
+  /** Updates a scheduledInterview Entity from a scheduledInterview object. */
+  public Entity scheduledInterviewToUpdateEntity(ScheduledInterview scheduledInterview) {
```
jalexanderqed: I don't think the description here is accurate since this does not actually update anything, just creates a new entity.

Related to that, I think `scheduledInterviewToEntity` or `scheduledInterviewToDatastoreEntity` would be a more appropriate name.

src/test/java/com/google/sps/data/DatastoreScheduledInterviewTest.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.List;
+import java.util.Optional;
+import org.junit.Assert;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DatastoreScheduledInterviewTest {
+
+  DatastoreScheduledInterviewDao tester = new DatastoreScheduledInterviewDao();
```
jalexanderqed: Please give this a more descriptive name than `tester`. Even just `dao` would be appropriate.

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
GrantComm: It's the method used in to update a stored scheduledInterview entity without losing to generated id. scheduledInterviewToEntity is already a method that creates a scheduledInterview entity with a generated id

src/test/java/com/google/sps/data/DatastoreScheduledInterviewTest.java
GrantComm: Done

src/test/java/com/google/sps/data/DatastoreScheduledInterviewTest.java
```diff
@@ -0,0 +1,188 @@
...
+    Optional<ScheduledInterview> expectedScheduledInterviewOptional =
+        Optional.of(expectedScheduledInterview);
+    Assert.assertEquals(expectedScheduledInterviewOptional, actualScheduledInterviewOptional);
+  }
+
+  // Tests retrieving a scheduledInterview that doesn't exist from Datastore.
+  @Test
+  public void failsGetScheduledInterview() {
+    Optional<ScheduledInterview> actual = dao.get(2);
+    Optional<ScheduledInterview> expected = Optional.empty();
```
jalexanderqed: This line is unnecessary. Just do `Assert.assertEquals(Optional.empty(), actual);`

src/main/java/com/google/sps/data/DatastoreScheduledInterviewDao.java
jalexanderqed: In that case, please rename to `scheduledInterviewToEntityForUpdate` or `scheduledInterviewToEntityWithId`. `scheduledInterviewToUpdateEntity` sounds too much like an "UpdateEntity" is a special kind of entity, which is not really accurate.

jalexanderqed (approved):

PR #99, opened 2020-07-15
Autofills email on registration form of logged in user

apluscs:


src/main/webapp/register.js
```diff
@@ -14,9 +14,19 @@
...
+  autofillEmail();
   prepareFormValidation();
 }
 
+// Autofills email on the registration form of the email of the logged in user.
+function autofillEmail() {
+  return fetch('/login')
+    .then(response => response.json())
+    .then((status) => {
+      $("#user-email").val(status.email);
```
meagancheese: Nit: if it doesn't effect functionality, switch to single quotes.

meagancheese (approved):
One nit, otherwise LGTM.

GrantComm (approved):
LGTM

src/main/webapp/register.js
```diff
@@ -14,9 +14,19 @@
 
 function onRegisterLoad() {
   supplyLogoutLink();
+  autofillEmail();
   prepareFormValidation();
```
acmcarther: This probably needs to be run after `autofillEmail()`. Since autofillEmail is asynchronous, you can't guarantee that it runs afterward unless you `then()` prepareFormValidation off of it

acmcarther (approved):
One comment

PR #101, opened 2020-07-15
Implements Fake Availability DAO.

meagancheese:


GrantComm (approved):
LGTM

src/test/java/com/google/sps/data/FakeAvailabilityDaoTest.java
```diff
@@ -0,0 +1,330 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.jupiter.api.Assertions;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.junit.Test;
+
+@RunWith(JUnit4.class)
+public class FakeAvailabilityDaoTest {
+  private final FakeAvailabilityDao tester = new FakeAvailabilityDao();
```
GrantComm: On another pr Jack want a more descriptive name such as "dao" instead of "tester"

GrantComm (changes requested):

src/test/java/com/google/sps/data/FakeAvailabilityDaoTest.java
meagancheese: Done

src/main/java/com/google/sps/data/FakeAvailabilityDao.java
```diff
@@ -0,0 +1,137 @@
...
+    List<Availability> userAvailability = getForUser(email);
+    return getInRange(userAvailability, minTime, maxTime);
+  }
+
+  /**
+   * Collects all Availabilities for the specified user within the specified time range that have
+   * been scheduled over, where minTime and maxTime are in milliseconds.
+   */
+  @Override
+  public List<Availability> getScheduledInRangeForUser(String email, long minTime, long maxTime) {
```
apluscs: lf we decide to delete this in the DatastoreDao we should delete it here too.

apluscs (approved):

src/main/java/com/google/sps/data/FakeAvailabilityDao.java
meagancheese: Deleted it in all instances (and relevant tests).

GrantComm (approved):
LGTM

src/main/java/com/google/sps/data/FakeAvailabilityDao.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+
+/** Mimics accessing Datastore to support managing Availability entities. */
+public class FakeAvailabilityDao implements AvailabilityDao {
+  // @param datastore the fake database. Made public because it will only be used
+  // while testing.
+  public LinkedHashMap<Long, Availability> datastore;
```
jalexanderqed: Naming this `datastore` is confusing, even though it is simulating the behavior of datastore. `storedObjects` or just `data` might work better.

jalexanderqed: This should be private since you can access the necessary information using other methods.

src/test/java/com/google/sps/data/FakeAvailabilityDaoTest.java
```diff
@@ -0,0 +1,271 @@
...
+          true);
+
+  // Checks that an Availability is stored in datastore.
+  @Test
+  public void createsAvailability() {
+    dao.create(availabilityOne);
+    List<Availability> storedAvailabilities = new ArrayList<Availability>(dao.datastore.values());
+    Availability expected =
+        Availability.create(
+            "user1@mail.com",
```
jalexanderqed: It would be clearer here if you used `availabilityOne.email()` and the other fields rather than repeating their values so that we know this is the same as the one provided above. As it is, the reader needs to realize that these are the same as those in `availabilityOne` and then go up and check that they are in fact the same.

src/test/java/com/google/sps/data/FakeAvailabilityDaoTest.java
```diff
@@ -0,0 +1,271 @@
...
+          new TimeRange(
+              Instant.parse("2020-07-07T22:30:00Z"), Instant.parse("2020-07-07T22:45:00Z")),
+          (long) -1,
+          true);
+
+  // Checks that an Availability is stored in datastore.
+  @Test
+  public void createsAvailability() {
+    dao.create(availabilityOne);
+    List<Availability> storedAvailabilities = new ArrayList<Availability>(dao.datastore.values());
```
jalexanderqed: This hash map is an implementation detail of the `FakeAvailabilityDao` class that should not be referenced in tests (I may have missed this in prior code review of fake DAOs). Your unit test should test the interface of the class, not its implementation. The contract this class provides is, effectively, "if you insert something, it will be available via the access methods", so testing only via the access methods is appropriate. This is distinct from the Datastore DAO objects, which need to guarantee "if you insert something, it will be created in Datastore," which is why you should assert on the contents of Datastore.

src/test/java/com/google/sps/data/FakeAvailabilityDaoTest.java
```diff
@@ -0,0 +1,271 @@
+// Copyright 2019 Google LLC
```
jalexanderqed: Discussed with Alex, and we agree that the fake DAOs do not need tests because they are themselves only used in tests. Please read my comments on this file anyway for future reference. I am sorry for all the extra work this caused.

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/FakeAvailabilityDao.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+
+/** Mimics accessing Datastore to support managing Availability entities. */
+public class FakeAvailabilityDao implements AvailabilityDao {
+  // @param datastore the fake database. Made public because it will only be used
```
jalexanderqed: I think @param is used for function arguments, not fields (please correct me if you have a counterexample).

jalexanderqed (approved):
Please update the PR name since this no longer adds tests.

PR #104, opened 2020-07-16
Redirected users to registration page if they are not registered yet

apluscs:


meagancheese (changes requested):
Is getUserIfRegistered() the best name for this method? Maybe it can't be helped because that is its main purpose on the registration page, but I think a separate method in a different file (like supplyLogoutLink is) would be better.

src/main/webapp/profile.js
```diff
@@ -13,8 +13,9 @@
 // limitations under the License.
 
 function onProfileLoad() {
-  supplyLogoutLink();  
-  getUserIfRegistered().then((person) => {
+  const loginInfo = getLoginInfo();
+  loginInfo.then(supplyLogoutLinkOrRedirectHome);  
+  loginInfo.then(getUserOrRedirectRegistration).then((person) => {
```
meagancheese, line 18: Does this function properly when the function returns nothing?

meagancheese (approved):
One question, but otherwise LGTM.

GrantComm (approved):

src/main/webapp/authentication.js
```diff
@@ -13,12 +13,27 @@
...
+// Returns a Person if they registered in the past. If not, redirect to  
+// registration page.
+function getUserOrRedirectRegistration(loginInfo){
+  return fetch(`/person?email=${loginInfo.email}`)
+    .then(response => {
+      if (response.redirected) {
+        window.location.href = response.url;
+        return;
+      }
+      return response.json();
```
jalexanderqed, line 37: Please confirm that this works manually. As we experienced during the meeting, `response.redirected` may not have the value you expect.

jalexanderqed (dismissed):
Feel free to send to Alex for final approval since I'm not sure I will get to it today.

src/main/webapp/authentication.js
apluscs: lt works when l checked it manually

acmcarther (approved):

PR #110, opened 2020-07-16
Connects the front and back ends of the Availability Page and updates test files.

meagancheese:
Let me know if there are any tests that seem to be missing.


src/main/java/com/google/sps/data/AvailabilityTimeSlot.java
```diff
@@ -57,6 +60,8 @@ static Builder builder() {
 
     abstract Builder setSelected(boolean selected);
 
+    abstract Builder setScheduled(boolean scheduled);
```
apluscs, line 65: What's the difference between selected vs scheduled?
Update: l know what they are but only after looking at some comment 7 files down. 
Could you put a comment here differentiating between the two?

src/main/java/com/google/sps/data/AvailabilityTimeSlotGenerator.java
```diff
@@ -119,7 +166,11 @@ static HoursAndMinutes create(int hour, int minute) {
 
       timeSlots.add(
           AvailabilityTimeSlot.create(
-              formattedUTCTime, formattedTime, formattedDate, getSelectedStatus(formattedUTCTime)));
+              formattedUTCTime,
+              formattedTime,
+              formattedDate,
+              getSelectedStatus(formattedUTCTime, userAvailabilityForWeek),
```
apluscs, line 180: These `get[x]Status` functions are time costly. Make a HashMap<Instant (startTime), Availability (a member of `userAvailabilityForWeek`)> so you can quickly look up an instant in `userAvailabilityForWeek`.
On the other hand, `userAvailabilityForWeek` won't be very large so l don't know even know if this matters. @jalexanderqed @acmcarther ?

src/main/java/com/google/sps/data/AvailabilityDao.java
```diff
@@ -22,10 +22,11 @@
  * AvailabilityDao includes the basic methods anything managing Availability entities must support.
  */
 public interface AvailabilityDao {
-  // Returns a list of all Availabilities ranging from minTime to maxTime of a user.
+  // Returns a sorted list of all Availabilities ranging from minTime to maxTime of a user.
```
apluscs: Nit: Please write how these are sorted (by startTime). And in all of these comments in the various files.

src/main/webapp/availability.js
```diff
@@ -20,10 +20,14 @@ function onAvailabilityLoad() {
 }
 
 // Toggles a tile from selected (green) to un-selected (white) and vice versa when clicked.
+// Scheduled tiles (red) remain unaffected.
 function toggleTile(tile) {
   let classList = tile.classList;
-  if (classList.contains('table-success') && classList.contains('selected-time-slot')) {
-    classList.remove('table-success', 'selected-time-slot');
+  if (classList.contains('selected-time-slot')) {
+    if (classList.contains('table-success')) {
```
apluscs: Q: how could something be a selected time slot but not be table-success?
Please write about what happens to scheduled slots in the comment of this function.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -47,6 +72,30 @@ public void doPut(HttpServletRequest request, HttpServletResponse response) thro
       response.sendError(400);
       return;
     }
-    // TODO: Use utcEncodings to generate Availabilities and store them in datastore.
+    UserService userService = UserServiceFactory.getUserService();
+    String email = userService.getCurrentUser().getEmail();
```
apluscs, line 76: Need to check if requestee == person logged in and return SC_UNAUTHORIZED (401)

src/test/java/com/google/sps/data/AvailabilityServletTest.java
```diff
@@ -67,6 +78,8 @@ public void validAvailabilityServletRequest() throws IOException {
   @Test
   public void invalidAvailabilityServletRequest() throws IOException {
```
apluscs: How is this invalid? Please put in a comment.
Also, since we're going to be getting availability too from the Servlet, we should specify in the test name if it's a get, put, or post.

apluscs: Add another test to check for what happens when someone is requestee for someone else (the 401 l mentioned above).

apluscs (changes requested):

src/main/webapp/availability.js
meagancheese: We are saying that both selected and scheduled availabilities are 'selected-time-slot's because that is the class that we use to send all availabilities on the page when the update button is clicked.

src/test/java/com/google/sps/data/AvailabilityServletTest.java
meagancheese: We are not actually getting availability from the servlet. This is because of how the AvailabilityTimeSlotGenerator and its corresponding jsp page are set up.

src/main/java/com/google/sps/data/AvailabilityTimeSlot.java
meagancheese: Done

src/main/java/com/google/sps/data/AvailabilityTimeSlotGenerator.java
meagancheese: Done (refactored to use HashMap).

src/main/java/com/google/sps/data/AvailabilityDao.java
meagancheese: Done

src/test/java/com/google/sps/data/AvailabilityServletTest.java
meagancheese: Added a comment

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
meagancheese: Not applicable because the request does not include info about the user.

src/test/java/com/google/sps/data/AvailabilityServletTest.java
meagancheese: Not applicable because the request does not include info about the user.

apluscs (approved):

GrantComm (approved):
LGTM

src/main/webapp/availability.js
jalexanderqed: I think it would be cleaner to have a separate tag to identify scheduled slots. This may already be how you're doing it, but it's hard to tell from this file?

jalexanderqed (changes requested):
One question in a response to Claire's comment.

jalexanderqed (approved):

PR #112, opened 2020-07-17
Uses Calendar API to get a Meet Link for a ScheduledInterview

apluscs:
CalendarAccess is the main thing


pom.xml
```diff
@@ -27,8 +27,53 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
-
+
```
GrantComm: Do we need this space?

src/main/java/com/google/sps/data/CalendarAccess.java
```diff
@@ -0,0 +1,93 @@
...
+import com.google.api.services.calendar.model.EventReminder;
+import com.google.api.services.calendar.model.Events;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
```
GrantComm: Are these used?

GrantComm (changes requested):
Just two questions

pom.xml
apluscs: Removed.

src/main/java/com/google/sps/data/CalendarAccess.java
apluscs: Removed some

GrantComm (approved):

src/main/java/com/google/sps/data/CalendarAccess.java
```diff
@@ -0,0 +1,91 @@
...
+      throws IOException, GeneralSecurityException {
+    Event event =
+        new Event()
+            .setSummary("Interview")
+            .setDescription(
+                "This event won't be shown to users, just used to \"reserve\" a Meet link.");
+
+    DateTime startDateTime = new DateTime(interview.when().start().toString());
+    EventDateTime start =
+        new EventDateTime().setDateTime(startDateTime).setTimeZone("America/Toronto");
```
meagancheese: Why is the timezone being set to this?

meagancheese (approved):
QQ about timezone, would like to see tests too.

meagancheese (changes requested):
We should really add a test file before it's called anywhere.

src/main/java/com/google/sps/data/CalendarAccess.java
apluscs: Part of example code, l'll remove it so it gets set to UTC

apluscs:
l have a test file  but left it out because it "interacts with the real world" - the same reason l left the email sending and secret fetching tests too.
Plus, it would break master because this code uses Secret Manager which only works in a App Engine environment.

src/main/java/com/google/sps/data/CalendarAccess.java
```diff
@@ -0,0 +1,91 @@
...
+    EventDateTime start = new EventDateTime().setDateTime(startDateTime);
+    System.out.println(start.getTimeZone());
+    event.setStart(start);
+
+    DateTime endDateTime = new DateTime(interview.when().end().toString());
+    EventDateTime end = new EventDateTime().setDateTime(endDateTime);
+    event.setEnd(end);
+
+    CreateConferenceRequest createRequest = new CreateConferenceRequest();
+    createRequest.setRequestId("randomstring");
```
apluscs: https://developers.google.com/resources/api-libraries/documentation/calendar/v3/java/latest/com/google/api/services/calendar/model/CreateConferenceRequest.html#setRequestId-java.lang.String-
Really don't know if we need to regenerate this each time. ln the test environment, l called this function twice in a row and 2 events popped up, even though the same requestId was being used.

src/main/java/com/google/sps/data/CalendarAccess.java
```diff
@@ -0,0 +1,91 @@
...
+import java.security.GeneralSecurityException;
+import java.util.Collections;
+
+// Handles all things Google Calendar (for now just getting a Meet link).
+public class CalendarAccess {
+  private Calendar service;
+  private static final String CALENDAR_ID = "info@jqed.dev";
+
+  // TODO: remember to write tests in the code that calls CalendarAccess() that handle what happens
+  // per each exception
```
meagancheese: If these files already exist but aren't safe for git, remove this TODO.

meagancheese (approved):

src/main/java/com/google/sps/data/CalendarAccess.java
apluscs: l thought we might make a stub of CalendarAccess that throws the same 4 exceptions. @acmcarther  ?

acmcarther: I'd typically recommend dependency injection here so that the constructor does not do non-trivial work. 

```java
public CalendarAccess(Calendar calendar);
```

Then, you can have a `public static` helper function which creates the calendar (and throws all of those exceptions)

```java
public static MakeCalendar(SecretFetcher secretFetcher) {
  String key = secretFetcher.getSecretValue("SERVICE_ACCT_KEY");
  return new Calendar.Builder(
      GoogleNetHttpTransport.newTrustedTransport(),
      JacksonFactory.getDefaultInstance(),
      GoogleCredential.fromStream(new ByteArrayInputStream(key.getBytes()))
            .createScoped(Collections.singletonList(CalendarScopes.CALENDAR)))
    .setApplicationName("Interview Me CalendarAccess")
    .build();
}
```

With the full usage being
```java
CalendarAccess calendarAccess = 
    new CalendarAccess(CalendarAccess.MakeCalendar(
        new SecretFetcher("interview-me-step-2020")));
```

That said, this is usually useful for testing, but I couldn't find a test Calendar class, so I don't know how useful this is. I guess a mock (from mockito) could be used...

src/main/java/com/google/sps/data/CalendarAccess.java
acmcarther: It says
> If an ID provided is the same as for the previous request, the request is ignored.
 
If the request is not idempotent when the same id is passed, then the API is lying to you. 

(probably best not to keep it as this fake string though)

acmcarther (changes requested):

acmcarther (changes requested):
This "omitting tests because they interact with the real service" thing is not good. As you've seen from chat with the email sender, this means that you're pushing the hassle of figuring out how to test around this class to the callers.

Tests are written for two reasons. First, to ensure to the author that the code works, and second to ensure that the code continues to work. The second point is actually much more important than the first.

My recommendation is in two parts:
1. Define an interface which your new class implements that others may supply a stub for in tests.
2. Separate out the event definition logic from the event creation logic, and test the event definition logic. This will probably take the shape of a function returning the event, marked `@VisibleForTesting`.

src/main/java/com/google/sps/data/CalendarAccess.java
```diff
@@ -0,0 +1,25 @@
...
+
+package com.google.sps.data;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+// Handles all things Calendar (for now just getting a Meet link).
+public interface CalendarAccess {
+
+  public String getMeetLink(ScheduledInterview interview)
```
acmcarther, line 25: Javadoc?

src/main/java/com/google/sps/data/FakeCalendarAccess.java
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+// A stub of CalendarAccess.
+public class FakeCalendarAccess implements CalendarAccess {
+
```
acmcarther: rm extra newline

src/main/java/com/google/sps/data/GoogleCalendarAccess.java
```diff
@@ -0,0 +1,95 @@
...
+
+  // TODO: remember to write tests in the code that calls CalendarAccess() that handle what happens
+  // per each exception
+  public GoogleCalendarAccess(Calendar service) {
+    this.service = service;
+  }
+
+  // Makes an event in the calendar CALENDAR_ID and returns the Meet Link associated with that
+  // event.
+  public static Event getEvent(ScheduledInterview interview) {
```
acmcarther: Make this package private (drop the public) and mark this `@VisibleForTesting`.

src/test/java/com/google/sps/data/GoogleCalendarAccessTest.java
```diff
@@ -0,0 +1,78 @@
...
+    helper.setUp();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void basic() throws IOException, GeneralSecurityException, URISyntaxException, Exception {
```
acmcarther: `throws Exception` is sufficient (the other types subclass Exception).

src/main/java/com/google/sps/data/GoogleCalendarAccess.java
acmcarther: Nit: This is more like `makeEvent`, not `getEvent`.

acmcarther (approved):

PR #116, opened 2020-07-21
Scheduling Process front and back end work.

meagancheese:
This code processes a search for interviews request and returns the possible times that interviewers are available for an interview during the next week. When a user clicks on a specific time slot, they are then presented with all of the possible company and job pairs that represent people that they could be interviewed by. When the user selects a company and job pair, an interview is scheduled with a randomly selected interviewer who is available during the specified time and has the selected company and job. Once an interview is scheduled, both parties availabilities during that time are marked as scheduled (updated).


src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,276 @@
...
+    // TODO: Decide how precise this range should be. Must start after or at now, when should end
+    // be?
+    Instant startOfRange = utcTime.toInstant();
+    Instant endOfRange = utcTime.toInstant().plus(6, ChronoUnit.DAYS);
+    List<Availability> availabilitiesInRange =
+        availabilityDao.getInRangeForAll(startOfRange, endOfRange);
+    List<PossibleInterviewSlot> possibleInterviews =
+        getPossibleInterviewSlots(availabilitiesInRange, startOfRange, endOfRange, timezoneOffset);
+    Set<String> dates = new HashSet<String>();
+    for (PossibleInterviewSlot pi : possibleInterviews) {
```
GrantComm: nit: change "pi" to "possibleInterview" for readability

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,276 @@
...
+      dates.add(pi.date());
+    }
+
+    List<ArrayList<PossibleInterviewSlot>> possibleInterviewsForWeek =
+        new ArrayList<ArrayList<PossibleInterviewSlot>>();
+    for (String date : dates) {
+      ArrayList<PossibleInterviewSlot> dayOfSlots = new ArrayList<PossibleInterviewSlot>();
+      for (PossibleInterviewSlot pi : possibleInterviews) {
```
GrantComm: Here too

src/main/webapp/possibleInterviewTimes.jsp
```diff
@@ -0,0 +1,40 @@
...
+  pageContext.setAttribute("list", list);
+%>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+
+<c:choose>
+  <c:when test= "${empty pageScope.list}">
+    <h2 style="text-align: center">No interviews are available at this time.</h2>
+  </c:when>
+  <c:otherwise>
+  <c:forEach items = "${pageScope.list}" var = "day">
```
GrantComm: Shouldn't the for each loop be tabbed once?

src/test/java/com/google/sps/servlets/ShowInterviewersServletTest.java
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  @Test
+  public void onlyReturnsInterviewersWhoAreAvailableForTheFullHour() throws IOException {
+    personDao.create(person1);
+    ShowInterviewersServlet servlet = new ShowInterviewersServlet();
+    servlet.init(availabilityDao, personDao);
+    helper.setEnvIsLoggedIn(true).setEnvEmail("person@gmail.com").setEnvAuthDomain("auth");
+
+    availabilityDao.create(
```
GrantComm: I see a lot of availabilities in the tests that use the same values. Would it be better to initialize them similar to the way you initialized values for users?

GrantComm (changes requested):

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/webapp/possibleInterviewTimes.jsp
meagancheese: Done

src/test/java/com/google/sps/servlets/ShowInterviewersServletTest.java
meagancheese: Done

GrantComm (approved):

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,276 @@
...
+  private List<PossibleInterviewSlot> getPossibleInterviewSlots(
+      List<Availability> allAvailabilities,
+      Instant startOfRange,
+      Instant endOfRange,
+      ZoneOffset timezoneOffset) {
+    Set<PossibleInterviewSlot> possibleInterviews = new HashSet<PossibleInterviewSlot>();
+    Set<String> interviewers = new HashSet<String>();
+    for (Availability avail : allAvailabilities) {
+      interviewers.add(avail.userId());
+    }
```
apluscs: l think this whole procedure to get the IDs of users whose availabilities fall within the given range can be made more simple/faster with Datastore's fancier query methods.
https://cloud.google.com/datastore/docs/concepts/queries#grouping
Grouping sounds like what you want. Would require another method to AvailabilityDAO.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,276 @@
...
+  // This method takes the timezoneOffsetMinutes int and converts it
+  // into a proper ZoneOffset instance.
+  private static ZoneOffset convertIntToOffset(int timezoneOffsetMinutes) {
+    return ZoneOffset.ofHoursMinutes((timezoneOffsetMinutes / 60), (timezoneOffsetMinutes % 60));
+  }
+
+  private List<PossibleInterviewSlot> getPossibleInterviewSlots(
+      List<Availability> allAvailabilities,
+      Instant startOfRange,
+      Instant endOfRange,
```
apluscs: To minimize params, could we replace this with a TimeRange instance?

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,276 @@
...
+      ArrayList<PossibleInterviewSlot> dayOfSlots = new ArrayList<PossibleInterviewSlot>();
+      for (PossibleInterviewSlot possibleInterview : possibleInterviews) {
+        if (date.equals(possibleInterview.date())) {
+          dayOfSlots.add(possibleInterview);
+        }
+      }
+      sortInterviews(dayOfSlots);
+      possibleInterviewsForWeek.add(dayOfSlots);
+    }
+    sortWeek(possibleInterviewsForWeek);
```
apluscs: `possibleInterviewsForWeek` is already sorted. You can just iterate over it, and when the date changes, append a new ArrayList<PossibleInterviewSlot> to `possibleInterviewsForWeek`
l feel like it's simpler and faster.

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,162 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String utc = request.getParameter("utc");
+    Instant startOfRange = Instant.parse(utc);
+    Instant endOfRange = Instant.parse(utc).plus(1, ChronoUnit.HOURS);
+    List<Availability> availabilitiesInRange =
+        availabilityDao.getInRangeForAll(startOfRange, endOfRange);
+    List<Person> possiblePeople =
+        getPossiblePeople(availabilitiesInRange, startOfRange, endOfRange);
+    Set<PossibleInterviewer> possibleInterviewers = getPossibleInterviewers(possiblePeople);
```
apluscs: This is confusing bc it seems like `possiblePeople` and `possibleInterviewers` are different, but `possibleInterviewers` is just the Set version of `possiblePeople`. `getPossiblePeople()` can just return a Set instead of a List and then converting it back into the Set it basically was originally.

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,162 @@
...
+  private Set<PossibleInterviewer> getPossibleInterviewers(List<Person> possiblePeople) {
+    Set<PossibleInterviewer> possibleInterviewers = new HashSet<PossibleInterviewer>();
+    for (Person person : possiblePeople) {
+      possibleInterviewers.add(personToPossibleInterviewer(person));
+    }
+
+    return possibleInterviewers;
+  }
+
+  boolean personHasPossibleInterviewSlot(String userId, Instant startOfRange, Instant endOfRange) {
```
apluscs: Q: `startOfRange` and `endOfRange` are exactly an hour apart right?
lf so, we can just check if `availabilities` after removing all `scheduledAvailability` has length == 4

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,162 @@
...
+    List<Availability> availabilities =
+        availabilityDao.getInRangeForUser(userId, startOfRange, endOfRange);
+    List<Availability> scheduledAvailability = new ArrayList<Availability>();
+    for (Availability avail : availabilities) {
+      if (avail.scheduled()) {
+        scheduledAvailability.add(avail);
+      }
+    }
+    availabilities.removeAll(scheduledAvailability);
```
apluscs: Shorten this with `removeIf()`
https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeIf-java.util.function.Predicate-

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
+    // TODO: Call a servlet to save this selection. Using interviewer.getAttribute('data-utc').
+    // TODO: Change this to data-id once refactor is merged
+    let requestObject = {
+      company: interviewer.getAttribute('data-company'),
+      job: interviewer.getAttribute('data-job'),
+      utc: interviewer.getAttribute('data-utc')
+    };
+    let requestBody = JSON.stringify(requestObject);
+    let request = new Request('/scheduled-interviews', {method: 'POST', body: requestBody});
+    fetch(request).then(unused => {window.location.replace('/scheduled-interviews.html');});
```
apluscs, line 62: Q: why `unused`? What's the meaning behind this?

apluscs (changes requested):

src/main/webapp/search-interview.js
meagancheese: Once an interview has been scheduled, redirect the user to the scheduled interviews page.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: They are different, notice the classes / class names. Person is different from PossibleInterviewer. I agree the naming of those variable is kind of unfortunate though. Open to suggestions.

meagancheese: Changed getPossibleInterviewers to peopleToPossibleInterviewers as discussed.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done. Not 100% sure it's faster but it has a more personalized dao method.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,264 @@
...
+    TimeRange range =
+        new TimeRange(utcTime.toInstant(), utcTime.toInstant().plus(6, ChronoUnit.DAYS));
+    List<PossibleInterviewSlot> possibleInterviews =
+        getPossibleInterviewSlots(range, timezoneOffset);
+
+    String date = "";
+    List<ArrayList<PossibleInterviewSlot>> possibleInterviewsForWeek =
+        new ArrayList<ArrayList<PossibleInterviewSlot>>();
+
+    if (possibleInterviews.size() != 0) {
```
apluscs: Nit: !possibleInterviews.isEmpty()
On second thought, just set `date = possibleInterviews.isEmpty() ? "" :  possibleInterviews.get(0).date();`

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,142 @@
...
+  boolean personHasPossibleInterviewSlot(String userId, TimeRange range) {
+    List<Availability> availabilities =
+        availabilityDao.getInRangeForUser(userId, range.start(), range.end());
+    availabilities.removeIf(avail -> avail.scheduled());
+    // If they have 4 unscheduled availabilities during the hour-long range, then they have
+    // a possible interview slot.
+    if (availabilities.size() == 4) {
+      return true;
+    }
+    return false;
```
apluscs: Nit: `return availabilities.size() == 4`

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,142 @@
...
+
+    return possibleInterviewers;
+  }
+
+  private Set<PossibleInterviewer> peopleToPossibleInterviewers(List<Person> possiblePeople) {
+    Set<PossibleInterviewer> possibleInterviewers = new HashSet<PossibleInterviewer>();
+    for (Person person : possiblePeople) {
+      possibleInterviewers.add(personToPossibleInterviewer(person));
+    }
+
```
apluscs: Nit: rm this line

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,142 @@
...
+    allInterviewers.remove(userId);
+
+    List<Person> possibleInterviewers = new ArrayList<Person>();
+    for (String interviewer : allInterviewers) {
+      if (personHasPossibleInterviewSlot(interviewer, range)) {
+        possibleInterviewers.add(personDao.get(interviewer).get());
+      }
+    }
+
```
apluscs: Nit: rm this line

src/main/webapp/search-interview.js
apluscs: Still confused. What's not used?

apluscs (changes requested):
Mostly nits.
Side note: this PR was really big, maybe next time you could split it up into the loading, then the showing, then the actual scheduling.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/webapp/search-interview.js
meagancheese: Whatever the fetch returns (typically the response I think) is unused here because we're not doing anything with info from the servlet apart from redirecting when the fetch is finished.

meagancheese:
Ready for another pass.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,263 @@
...
+              .get(i + 2)
+              .when()
+              .end()
+              .equals(availabilities.get(i + 3).when().start())) {
+            possibleInterviewSlotsForPerson.add(
+                PossibleInterviewSlot.create(
+                    availabilities.get(i).when().start().toString(),
+                    getDate(availabilities.get(i).when().start(), timezoneOffset),
+                    getTime(availabilities.get(i).when().start(), timezoneOffset)));
+          }
```
apluscs: Since `availabilities` is sorted, we can just check if the availability [i + 3] is at current time + 45 minutes

apluscs (approved):
One thing

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Good call, done.

src/main/java/com/google/sps/data/Availability.java
```diff
@@ -42,6 +42,11 @@ public Availability withId(long id) {
     return toBuilder().setId(id).build();
   }
 
+  // Returns a new Availability with the old information with an updated scheduled status.
+  public Availability scheduledStatus(boolean scheduled) {
```
acmcarther: Follow naming scheme of similar function.

(`withScheduled`)

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -171,4 +174,31 @@ public void deleteInRangeForUser(String userId, Instant minTime, Instant maxTime
             .addSort("startTime", SortDirection.ASCENDING);
     return datastore.prepare(availQuery).asList(FetchOptions.Builder.withDefaults());
   }
+
+  // Returns the ids of all users that have availabilities within the specified time range.
+  public Set<String> getUsersAvailableInRange(Instant minTime, Instant maxTime) {
+    Filter minFilter =
```
acmcarther: These field names are probably more useful if they're named after the property they're filtering, instead of the way they're filtering that property. Same applies for below.

So, `startTimeFilter` here.

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -171,4 +174,31 @@ public void deleteInRangeForUser(String userId, Instant minTime, Instant maxTime
...
+            "startTime", FilterOperator.GREATER_THAN_OR_EQUAL, minTime.toEpochMilli());
+    // Queries can only perform inequality filters on one parameter, and so instead
+    // of using endTime for the maxFilter, startTime is used and the maxTime has 15
+    // minutes subtracted from it to be equal to the latest possible startTime.
+    Filter maxFilter =
+        new FilterPredicate(
+            "startTime",
+            FilterOperator.LESS_THAN_OR_EQUAL,
+            maxTime.minus(15, ChronoUnit.MINUTES).toEpochMilli());
+    CompositeFilter compFilter = CompositeFilterOperator.and(minFilter, maxFilter);
```
acmcarther: `startAndEndFilter`

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+    Set<String> interviewers = availabilityDao.getUsersAvailableInRange(range.start(), range.end());
+    Set<PossibleInterviewSlot> possibleInterviews = new HashSet<PossibleInterviewSlot>();
+
+    // We don't want to schedule an interview for a user with themself, so we are removing
+    // the current user's id from the list.
+    UserService userService = UserServiceFactory.getUserService();
+    String userEmail = userService.getCurrentUser().getEmail();
+    String userId = userService.getCurrentUser().getUserId();
+
+    // Since UserId does not have a valid Mock, if the id is null (as when testing), it will be
```
acmcarther: I know this comment has been copy pasted all over the code base (yuck) but it doesn't actually make sense.

The real cause is that "Users returned from the LocalUserService (in tests) do not have userIds".

src/main/java/com/google/sps/data/InterviewPostRequest.java
```diff
@@ -0,0 +1,51 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.ArrayList;
+
+public class InterviewPostRequest {
```
acmcarther, line 24: Javadoc?

src/main/java/com/google/sps/data/InterviewPostRequest.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public boolean allFieldsPopulated() {
+    return !(company == null || job == null || utc == null);
+  }
+
+  public String toString() {
+    return String.format(
+        "%s= %s:%s, %s:%s, %s:%s",
+        "InterviewPostRequest", "company", company, "job", job, "utc", utc);
```
acmcarther: There's not really any reason (that I know of) to substitute parameters which have known names. Inline those into the format string.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+
+  @Override
+  public void init() {
+    init(new DatastoreAvailabilityDao(), new DatastoreScheduledInterviewDao(), Instant.now());
+  }
+
+  public void init(
+      AvailabilityDao availabilityDao,
+      ScheduledInterviewDao scheduledInterviewDao,
+      Instant instant) {
```
acmcarther: What is this `instant` supposed to be?

(That question should be answered by a good parameter name)

Fix the class field as well.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int timezoneOffsetMinutes = Integer.parseInt(request.getParameter("timeZoneOffset"));
+    Preconditions.checkArgument(
+        Math.abs(timezoneOffsetMinutes) <= 720,
+        "Offset greater than 720 minutes (12 hours): %s",
+        timezoneOffsetMinutes);
+    ZoneOffset timezoneOffset = convertIntToOffset(timezoneOffsetMinutes);
+    ZonedDateTime day = generateDay(instant, timezoneOffset);
+    ZonedDateTime utcTime = day.withZoneSameInstant(ZoneOffset.UTC);
+    // TODO: Decide how precise this range should be. Must start after or at now, when should end
```
acmcarther: ?????

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+    this.availabilityDao = availabilityDao;
+    this.scheduledInterviewDao = scheduledInterviewDao;
+    this.instant = instant;
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int timezoneOffsetMinutes = Integer.parseInt(request.getParameter("timeZoneOffset"));
+    Preconditions.checkArgument(
+        Math.abs(timezoneOffsetMinutes) <= 720,
```
acmcarther: Declare this constant and reference it in the error messagel.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+        "Offset greater than 720 minutes (12 hours): %s",
+        timezoneOffsetMinutes);
+    ZoneOffset timezoneOffset = convertIntToOffset(timezoneOffsetMinutes);
+    ZonedDateTime day = generateDay(instant, timezoneOffset);
+    ZonedDateTime utcTime = day.withZoneSameInstant(ZoneOffset.UTC);
+    // TODO: Decide how precise this range should be. Must start after or at now, when should end
+    // be?
+    TimeRange range =
```
acmcarther: "range"?

interviewTimeRange? or interviewTime?

(I think "interviewRange" doesn't give any intuition, so i skipped that option)

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+        new TimeRange(utcTime.toInstant(), utcTime.toInstant().plus(6, ChronoUnit.DAYS));
+    List<PossibleInterviewSlot> possibleInterviews =
+        getPossibleInterviewSlots(range, timezoneOffset);
+
+    String date = possibleInterviews.isEmpty() ? "" : possibleInterviews.get(0).date();
```
acmcarther, line 95: It sounds like possibleInterviews.isEmpty() should return an error or otherwise be handled, but I don't see it handled at all.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+
+  // This method takes the timezoneOffsetMinutes int and converts it
+  // into a proper ZoneOffset instance.
+  private static ZoneOffset convertIntToOffset(int timezoneOffsetMinutes) {
+    return ZoneOffset.ofHoursMinutes((timezoneOffsetMinutes / 60), (timezoneOffsetMinutes % 60));
+  }
+
+  private List<PossibleInterviewSlot> getPossibleInterviewSlots(
+      TimeRange range, ZoneOffset timezoneOffset) {
+
```
acmcarther: Remove blank line.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  // Uses an Instant and a timezoneOffset to create a ZonedDateTime instance.
+  private static ZonedDateTime generateDay(Instant instant, ZoneOffset timezoneOffset) {
+    return instant.atZone(ZoneId.ofOffset("UTC", timezoneOffset));
+  }
+
+  // This method takes the timezoneOffsetMinutes int and converts it
```
acmcarther: Nit: Drop "this method"

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+    UserService userService = UserServiceFactory.getUserService();
+    String userEmail = userService.getCurrentUser().getEmail();
+    String userId = userService.getCurrentUser().getUserId();
+
+    // Since UserId does not have a valid Mock, if the id is null (as when testing), it will be
+    // replaced with this hashcode.
+    if (userId == null) {
+      userId = String.format("%d", userEmail.hashCode());
+    }
+
```
acmcarther: Drop this newline and L140 so that it is obvious that your comment on L135 applies all the way to this line.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+          getPossibleInterviewSlotsForPerson(interviewer, range, timezoneOffset));
+    }
+
+    // We need to check that the person looking to schedule is not already scheduled during any of
+    // the proposed times.
+
+    List<ScheduledInterview> userScheduledInterviews =
+        scheduledInterviewDao.getScheduledInterviewsInRangeForUser(
+            userId, range.start(), range.end());
+
```
acmcarther: Nit: drop most of these blank lines (up until L172, I think).

The blank lines are intended to help readers understand logical chunks.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+  private List<PossibleInterviewSlot> getPossibleInterviewSlotsForPerson(
+      String userId, TimeRange range, ZoneOffset timezoneOffset) {
+    List<Availability> availabilities =
+        availabilityDao.getInRangeForUser(userId, range.start(), range.end());
+    availabilities.removeIf(avail -> avail.scheduled());
+
+    List<PossibleInterviewSlot> possibleInterviewSlotsForPerson =
+        new ArrayList<PossibleInterviewSlot>();
+    for (int i = 0; i < availabilities.size() - 3; i++) {
```
acmcarther: This `3` needs to be made a constant.

No magic numbers please.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+      if (availabilities
```
acmcarther: Can this condition be assigned into a boolean instead of turning into a spooky long call chain?

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+    if (hour > 12) {
+      standardHour = hour - 12;
+    }
+    return String.format("%d:%02d %s", standardHour, minute, hour < 12 ? "AM" : "PM");
+  }
+
+  private void sortInterviews(List<PossibleInterviewSlot> possibleInterviewSlots) {
+    possibleInterviewSlots.sort(
+        (PossibleInterviewSlot p1, PossibleInterviewSlot p2) -> {
+          if (Instant.parse(p1.utcEncoding()).equals(Instant.parse(p2.utcEncoding()))) {
```
acmcarther: Best not to parse these values a bunch of times. 

Declare them in advance and then reference them.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
```diff
@@ -0,0 +1,256 @@
...
+            return -1;
+          }
+          return 1;
+        });
+  }
+
+  private void sortWeek(List<ArrayList<PossibleInterviewSlot>> possibleInterviewsForWeek) {
+    possibleInterviewsForWeek.sort(
+        (ArrayList<PossibleInterviewSlot> d1, ArrayList<PossibleInterviewSlot> d2) -> {
+          if (Instant.parse(d1.get(0).utcEncoding())
```
acmcarther: Same here.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -83,36 +101,87 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String requestedInterviewerId = request.getParameter("interviewer");
-    String requestedIntervieweeId = request.getParameter("interviewee");
-    String userEmail = UserServiceFactory.getUserService().getCurrentUser().getEmail();
-    String userId = UserServiceFactory.getUserService().getCurrentUser().getUserId();
+    String intervieweeEmail = userService.getCurrentUser().getEmail();
+    String intervieweeId = userService.getCurrentUser().getUserId();
+
     // Since UserId does not have a valid Mock, if the id is null (as when testing), it will be
```
acmcarther: Mentioned that this comment is not correct above, can you fix the references to this in this file too to match?

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -83,36 +101,87 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-          ScheduledInterview.create(
-              defaultKey,
-              new TimeRange(
-                  Instant.parse(request.getParameter("startTime")),
-                  Instant.parse(request.getParameter("endTime"))),
-              requestedInterviewerId,
-              requestedIntervieweeId);
-      scheduledInterviewDao.create(scheduledInterview);
+    if (!postRequest.allFieldsPopulated()) {
+      response.sendError(400);
```
acmcarther, line 108: Is it possible to log which fields are missing?

Actually, why are we doing this at all?

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -83,36 +101,87 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      userId = String.format("%d", userEmail.hashCode());
+    if (intervieweeId == null) {
+      intervieweeId = String.format("%d", intervieweeEmail.hashCode());
     }
-    // The default key for a scheduledInterview being stored in datastore
-    long defaultKey = -1;
-    if ((!requestedInterviewerId.equals(userId)) && (!requestedIntervieweeId.equals(userId))) {
-      response.sendError(HttpServletResponse.SC_UNAUTHORIZED);
+
+    // TODO: Put this null id check in a separate method to call from both Get and Post.
```
acmcarther: ???

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  public void init(AvailabilityDao availabilityDao, PersonDao personDao) {
+    this.availabilityDao = availabilityDao;
+    this.personDao = personDao;
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String utc = request.getParameter("utc");
```
acmcarther: What is this?

"utc" is a specific time zone, but we've called a parameter "utc". Is this a timestamp in utc?

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,137 @@
...
+    TimeRange range =
```
acmcarther: Variable names should convey semantics of what they contain, when possible. Often the type name is adequate, but it often is not.

Maybe "interviewTimeRange"?

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,137 @@
...
+      allInterviewers.add(avail.userId());
+    }
+
+    // We don't want to schedule an interview for a user with themself, so we are removing
+    // the current user's id from the list.
+    UserService userService = UserServiceFactory.getUserService();
+    String userEmail = userService.getCurrentUser().getEmail();
+    String userId = userService.getCurrentUser().getUserId();
+
+    // Since UserId does not have a valid Mock, if the id is null (as when testing), it will be
```
acmcarther: Fix this incorrect comment too.

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
```diff
@@ -0,0 +1,137 @@
...
+    // replaced with this hashcode.
+    if (userId == null) {
+      userId = String.format("%d", userEmail.hashCode());
+    }
+
```
acmcarther: Remove spaces to keep the logical chunk together (this chunk being related to dropping the user)

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
-// Should query Datastore for appropriate interviews and render them on the
+// Queries Datastore for available interview times and renders them on the
 // page.
 function loadInterviews() {
-  // TODO: fetch and get all interviews instead of unhiding something hidden
   const searchResultsDiv = document.getElementById("search-results");
   searchResultsDiv.removeAttribute("hidden");
+
+  fetch(`/load-interviews?timeZoneOffset=${browserTimezoneOffset()}`)
+  .then(response => response.text())
```
acmcarther, line 28: Fix indentation.

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
 }
 
 // Confirms interview selection with user and sends this selection to Datastore 
 // if confirmed.
 function selectInterview(interviewer) {
   if (confirm(
-    // TOOD: fill in these times dynamically from ids in the .jsp file.
-      `You selected: Sunday 7/5 from 6:30 PM - 7:30 PM with a ` +
-      `${interviewer.getAttribute("data-company")} ` +
+      `You selected: ${interviewer.getAttribute('data-date')} from ` +
```
acmcarther: Declare a variable

(generally, inlining long stuff into a conditional (if) looks awful)

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
-      `You have scheduled an interview on Sunday 7/5 from 6:30 PM - 7:30 ` +
-      `PM with a ${interviewer.getAttribute('data-company')} ` +
+      `You have scheduled an interview on ${interviewer.getAttribute('data-date')}` +
+      ` from ${interviewer.getAttribute('data-time')} ` +
+      `with a ${interviewer.getAttribute('data-company')} ` +
       `${interviewer.getAttribute('data-job')}. Check your email for more ` +
       `information.`);
-    // TODO: Call a servlet to save this selection.
-    location.reload();
+    // TODO: Call a servlet to save this selection. Using interviewer.getAttribute('data-utc').
```
acmcarther: ?????

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
+    // TODO: Change this to data-id once refactor is merged
```
acmcarther: ?????

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
-      // TODO: format the date as June 5, 2020 not 7/5/2020 (more readable).
-      const date = selectButton.getAttribute("id");
-      const time = document.getElementById(date + '-options').value;
-      $('#modal-title').text(`Interviewers Information for ${date} at ${time}`);
+  const date = selectButton.getAttribute('data-date');
+  const time = document.getElementById(date).innerText;
+  const reformattedTime = time.replace('-', 'to');
+  const utc = document.getElementById(date).value;
+  fetch(`/show-interviewers?utc=${utc}&date=${date}&time=${reformattedTime}`)
+  .then(response => response.text())
```
acmcarther, line 74: fix indentation.

src/test/java/com/google/sps/data/DatastoreAvailabilityDaoTest.java
```diff
@@ -78,6 +80,14 @@
           (long) -1,
           true);
 
+  private final Availability availabilityFive =
+      Availability.create(
+          "user3",
+          new TimeRange(
+              Instant.parse("2020-07-07T22:30:00Z"), Instant.parse("2020-07-07T22:45:00Z")),
+          (long) -1,
```
acmcarther: parameter comment, so readers can tell what this is supposed to be.

```
/*id=*/-1,
```
Do this for these two params

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
```diff
@@ -54,143 +54,218 @@
...
-  }
-
-  @After
-  public void tearDown() {
-    helper.tearDown();
-  }
-
-  private final Person interviewer =
+  private final Person person1 =
       Person.create(
```
acmcarther, line 61: This really is way too many sequential string params, but I guess we can't change that in this PR.

These get really easy to mix up once there's more than like two of them in a row.

Just don't do this in future code you write.

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
```diff
@@ -54,143 +54,218 @@
...
-          "Subject",
+          emailToId("user2@mail"),
+          "user2@mail",
+          "User",
+          "Test",
           "Google",
           "SWE",
           "linkedIn",
           EnumSet.noneOf(Job.class));
+  private final Availability person2Avail1 =
```
acmcarther: These names are pretty bad.

Is there an interesting property that these instances can each be named after?

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
```diff
@@ -219,18 +294,158 @@ public void orderedScheduledInterviewServletGetRequest() throws IOException {
...
+    ScheduledInterviewServlet scheduledInterviewServlet = new ScheduledInterviewServlet();
+    scheduledInterviewServlet.init(scheduledInterviewDao, availabilityDao, personDao);
+    helper.setEnvIsLoggedIn(true).setEnvEmail("user@company.org").setEnvAuthDomain("auth");
+    MockHttpServletRequest postRequest = new MockHttpServletRequest();
+    MockHttpServletResponse postResponse = new MockHttpServletResponse();
+    String jsonString = "{\"company\":\"Google\",\"job\":\"SWE\",\"utc\":\"2020-07-20T12:45:00Z\"}";
+    postRequest.setContent(jsonString.getBytes(StandardCharsets.UTF_8));
+
+    scheduledInterviewServlet.doPost(postRequest, postResponse);
+
```
acmcarther: These newlines seem arbitrary.

acmcarther (changes requested):
Didn't really review the tests yet

src/main/java/com/google/sps/data/Availability.java
meagancheese: Done

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: Done

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/data/InterviewPostRequest.java
meagancheese: Added.

src/main/java/com/google/sps/data/InterviewPostRequest.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Changed to currentTime.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Fixed

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: I think it's already included with the next line "Offset greater than 720 minutes (12 hours): %d", timezoneOffsetMinutes);
If there's something else that needs to be done, let me know.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Changed to interviewSearchTimeRange

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: On the possibleInterviewTimes jsp page I return "No interviews are available at this time." It's possible for there to be no interviews available, so I wasn't throwing an error but explaining to the user the current situation.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done. Deleted this function because I'm no longer using it.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
meagancheese: Handled.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
meagancheese: This is because the Gson.toJson() returns a null field if one of the fields is malformed. Either the name of the field is wrong or the input is bad. Basically an additional catch to see if we have a bad request.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
meagancheese: Done. Created a method getUserId() that handles the null case.

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: I've changed it to utcStartTime

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ShowInterviewersServlet.java
meagancheese: Done

src/main/webapp/search-interview.js
meagancheese: Done

src/main/webapp/search-interview.js
meagancheese: Done

src/main/webapp/search-interview.js
meagancheese: Removed

src/main/webapp/search-interview.js
meagancheese: Removed

src/main/webapp/search-interview.js
meagancheese: Done

src/test/java/com/google/sps/data/DatastoreAvailabilityDaoTest.java
meagancheese: Done for userId and id.

meagancheese: Let me know if it's necessary for the scheduled field as well.

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
meagancheese: Removed most blank lines.

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
meagancheese: Honestly the only interesting property is the person. All 3 availability sets are the same apart from the userId. I started renaming and realized that it didn't make much sense.

src/test/java/com/google/sps/data/DatastoreAvailabilityDaoTest.java
acmcarther: Yes.

The point is that the bare boolean or int or similar are not self evident from the call site.

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
acmcarther: If these don't have a distinct identifying property then they likely don't benefit at all from being factored out here. I'd argue that they're actually harming readability, since it's not obvious at all in the usage sites what specifically these values represent.

The tradeoffs in tests are slightly different than regular code. In tests, the priority is that the code is self-evident, and often that means that you should inline and permit additional duplication where it enhances clarity.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
acmcarther: Declare this constant and reference it in the error message.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
acmcarther: This hasn't been pushed I guess.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
acmcarther: The constant is 720, if that wasn't clear.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
acmcarther: Alright I guess.

(The red flag here is assigning what is ostensibly an encoded date to empty string, but no matter I guess. We're time crunched after all)

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
acmcarther: Ack.

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
acmcarther: Marking resolved.

src/main/java/com/google/sps/data/PossibleInterviewSlot.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * A PossibleInterviewSlot represents a 1 hour chunk of time with a utcEncoding and readable date
+ * and time strings.
+ */
+@AutoValue
+public abstract class PossibleInterviewSlot {
+  public abstract String utcEncoding();
```
acmcarther, line 25: Is this an encoded timestamp in utc?

UTC encoding sounds like "the description of how a utc timestamp is encoded", not "an encoded timestamp"

acmcarther: I can see that there's a bunch of existing code that uses this unfortunate verbiage, so it might be onerous to fix this.

Oh well.

src/main/webapp/possibleInterviewers.jsp
```diff
@@ -0,0 +1,38 @@
...
+     </tr>
+  </thead>
+  <tbody>
+    <c:forEach items = "${pageScope.set}" var = "interviewer">
+      <tr>
+        <td>${interviewer.company()}</td>
+        <td>${interviewer.job()}</td>
+        <td>
+          <button type="button" class="btn btn-primary" 
+          data-company="${interviewer.company()}" data-job="${interviewer.job()}"
```
acmcarther: Fix indentation

src/test/java/com/google/sps/data/DatastoreAvailabilityDaoTest.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Done

src/main/java/com/google/sps/servlets/LoadInterviewsServlet.java
meagancheese: Sadly github doesn't update comment code snippets.

src/main/webapp/possibleInterviewers.jsp
meagancheese: Done

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
meagancheese: I renamed the Person instances which I think helps bring some clarity.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -83,36 +96,81 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+              Instant.parse(utcStartTime), Instant.parse(utcStartTime).plus(1, ChronoUnit.HOURS));
     } catch (DateTimeParseException e) {
       response.sendError(400, e.getMessage());
       return;
     }
+
+    List<Availability> availabilitiesInRange =
+        availabilityDao.getInRangeForAll(range.start(), range.end());
+    List<Person> allAvailableInterviewers =
+        interviewerServlet.getPossiblePeople(availabilitiesInRange, range);
```
jalexanderqed: Calling a method on an instance of another servlet is kind of weird, as is having one servlet own an instance of another servlet. If possible, I think you should make getPossiblePeople a static method and call `ShowInterviewersServlet.getPossiblePeople()`

jalexanderqed (changes requested):

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
meagancheese: Done. Involved passing around some daos, but it works.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -46,29 +52,36 @@
 public class ScheduledInterviewServlet extends HttpServlet {
 
   private ScheduledInterviewDao scheduledInterviewDao;
+  private AvailabilityDao availabilityDao;
   private PersonDao personDao;
+  private final UserService userService = UserServiceFactory.getUserService();
+  // private ShowInterviewersServlet interviewerServlet = new ShowInterviewersServlet();
```
jalexanderqed: Please take this line out entirely.

jalexanderqed (approved):

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
jalexanderqed: Also a comment one in init()

meagancheese: Oops, will do.

meagancheese: Done

src/main/java/com/google/sps/data/AvailabilityTimeSlotGenerator.java
```diff
@@ -77,16 +77,20 @@ static HoursAndMinutes create(int hour, int minute) {
    */
   public static List<List<AvailabilityTimeSlot>> timeSlotsForWeek(
       Instant instant, int timezoneOffsetMinutes, AvailabilityDao availabilityDao) {
+    int maxTimezoneOffsetMinutes = 720;
```
acmcarther: nit: class constant, not a regular variable.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -83,36 +94,82 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      response.sendError(400);
       return;
     }
+
+    String interviewerCompany = postRequest.getCompany();
+    String interviewerJob = postRequest.getJob();
+    String utcStartTime = postRequest.getUtcStartTime();
+
+    TimeRange range;
+
```
acmcarther, line 116: nit: rm extra newline

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -83,36 +94,82 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    List<String> possibleInterviewers =
+        getPossibleInterviewerIds(allAvailableInterviewers, interviewerCompany, interviewerJob);
+
+    int randomNumber = (int) (Math.random() * possibleInterviewers.size());
+    String interviewerId = possibleInterviewers.get(randomNumber);
+
+    scheduledInterviewDao.create(
+        ScheduledInterview.create(-1, range, interviewerId, intervieweeId));
+
+    // Since an interview was scheduled, both parties' availabilities must be updated
```
acmcarther, line 140: This is not something you're going to be able to fix with the time remaining, but I think this represents a possible race condition.

Basically, you create a could scheduled interview for two interviewees and the same interviewer at the same time. Both ScheduledInterview would get created, but only one interview can actually impact the availability.

src/main/webapp/possibleInterviewers.jsp
```diff
@@ -0,0 +1,38 @@
+<%@ page import="java.util.Set" %>
+<%@ page import="com.google.sps.data.PossibleInterviewer" %>
+<%
+  Set<PossibleInterviewer> set = (Set<PossibleInterviewer>) request.getAttribute("interviewers");
+  pageContext.setAttribute("set", set);
+  String utc = request.getParameter("utcStartTime");
```
acmcarther: Any reason not to inline all of these into pageContext.setAttribute?

src/main/webapp/search-interview.html
```diff
@@ -44,86 +44,18 @@ <h5 class="text-center lead">Search for possible times in which you will be inte
       <form>
         <div class="form-group">
           <!-- TODO: add parameters for this form. -->
```
acmcarther, line 47: Drop this todo?

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
-// Should query Datastore for appropriate interviews and render them on the
+// Queries Datastore for available interview times and renders them on the
 // page.
 function loadInterviews() {
-  // TODO: fetch and get all interviews instead of unhiding something hidden
   const searchResultsDiv = document.getElementById("search-results");
   searchResultsDiv.removeAttribute("hidden");
+
+  fetch(`/load-interviews?timeZoneOffset=${browserTimezoneOffset()}`)
+    .then(response => response.text())
```
acmcarther: Fix indentation (`.then`s should be aligned here)

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
+      .then(interviewTimes => {
+        interviewTimesDiv().innerHTML = interviewTimes;
+      });
+}
+
+function interviewTimesDiv() {
+  return document.getElementById('interview-times-container');
+}
+
+function browserTimezoneOffset() {
```
acmcarther: This is duplicated from https://github.com/googleinterns/InterviewMe/blob/2d33d23e54a4f731c9f2c474ce1a6aacfbef9557/src/main/webapp/availability.js#L43-L46

Please deduplicate all duplicated code by moving it into a common file where possible.

src/main/webapp/search-interview.js
```diff
@@ -18,42 +18,67 @@ function onSearchInterviewLoad() {
...
-      const time = document.getElementById(date + '-options').value;
-      $('#modal-title').text(`Interviewers Information for ${date} at ${time}`);
-    });
+  const date = selectButton.getAttribute('data-date');
+  const time = document.getElementById(date).innerText;
+  const reformattedTime = time.replace('-', 'to');
+  const utc = document.getElementById(date).value;
+  fetch(`/show-interviewers?utcStartTime=${utc}&date=${date}&time=${reformattedTime}`)
+    .then(response => response.text())
+      .then(interviewers => {
```
acmcarther: fix indentation.

(why is the indentation on all of these messed up?)

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
```diff
@@ -54,184 +54,371 @@
...
-    personDao = new FakePersonDao();
-  }
-
-  @After
-  public void tearDown() {
-    helper.tearDown();
-  }
-
-  private final Person interviewer =
+  private final Person GoogleSWE1 =
```
acmcarther: Fix variable capitalization (here and below)

acmcarther (changes requested):
~I think I'll make one more pass tomorrow morning and then it's all up to Jack~
Nevermind, I'm the last reviewer.

acmcarther (approved):

src/main/java/com/google/sps/data/AvailabilityTimeSlotGenerator.java
meagancheese: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
meagancheese: Done

src/main/webapp/possibleInterviewers.jsp
meagancheese: Inlined.

src/main/webapp/search-interview.html
meagancheese: No, this is some of the qualifications and shadow work.

src/main/webapp/search-interview.js
meagancheese: Done

src/main/webapp/search-interview.js
meagancheese: Done

src/main/webapp/search-interview.js
meagancheese: Done

src/test/java/com/google/sps/data/ScheduledInterviewServletTest.java
meagancheese: Done

PR #117, opened 2020-07-21
Implements JSP page for ScheduledInterviews

GrantComm:
Implements a JSP page that displays the scheduled interviews for a particular user

### The scheduled interviews page with a scheduled interview looks like this (EST): 
Screenshot 2020-07-23 at 11 41 23 AM

### DIFF
--- a/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
+++ b/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
@@ -18,16 +18,16 @@ package com.google.sps.data;
 public class ScheduledInterviewRequest {
   private long id;
   private String dateString;
-  private String interviewerId;
-  private String intervieweeId;
+  private String interviewer;
+  private String interviewee;
   private String role;
 
   public ScheduledInterviewRequest(
-      long id, String dateString, String interviewerId, String intervieweeId, String role) {
+      long id, String dateString, String interviewer, String interviewee, String role) {
     this.id = id;
     this.dateString = dateString;
-    this.interviewerId = interviewerId;
-    this.intervieweeId = intervieweeId;
+    this.interviewer = interviewer;
+    this.interviewee = interviewee;
     this.role = role;
   }
 
@@ -39,12 +39,12 @@ public class ScheduledInterviewRequest {
     return dateString;
   }
 
-  public String getInterviewerId() {
-    return interviewerId;
+  public String getInterviewer() {
+    return interviewer;
   }
 
-  public String getIntervieweeId() {
-    return intervieweeId;
+  public String getInterviewee() {
+    return interviewee;
   }
 
   public String getRole() {
@@ -59,8 +59,8 @@ public class ScheduledInterviewRequest {
       ScheduledInterviewRequest that = (ScheduledInterviewRequest) o;
       return this.getId() == that.getId()
           && this.getDateString().equals(that.getDateString())
-          && this.getInterviewerId().equals(that.getInterviewerId())
-          && this.getIntervieweeId().equals(that.getIntervieweeId())
+          && this.getInterviewer().equals(that.getInterviewer())
+          && this.getInterviewee().equals(that.getInterviewee())
           && this.getRole().equals(that.getRole());
     }
     return false;
@@ -73,10 +73,10 @@ public class ScheduledInterviewRequest {
         id,
         "dateString",
         dateString,
-        "interviewerId",
-        interviewerId,
-        "intervieweeId",
-        intervieweeId,
+        "interviewer",
+        interviewer,
+        "interviewee",
+        interviewee,
         "role",
         role);
   }
--- a/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
+++ b/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
@@ -14,7 +14,7 @@
 
 package com.google.sps.data;
 
-/** Represents the data sent in a put or post request to the Person Servlet. */
+/** Represents the data sent from a get request to the ScheduledInterview Servlet. */
 public class ScheduledInterviewRequest {
   private long id;
   private String dateString;
--- a/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
+++ b/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
@@ -17,15 +17,15 @@ package com.google.sps.data;
 /** Represents the data sent in a put or post request to the Person Servlet. */
 public class ScheduledInterviewRequest {
   private long id;
-  private TimeRange when;
+  private String dateString;
   private String interviewerId;
   private String intervieweeId;
   private String role;
 
   public ScheduledInterviewRequest(
-      long id, TimeRange when, String interviewerId, String intervieweeId) {
+      long id, String dateString, String interviewerId, String intervieweeId, String role) {
     this.id = id;
-    this.when = when;
+    this.dateString = dateString;
     this.interviewerId = interviewerId;
     this.intervieweeId = intervieweeId;
     this.role = role;
@@ -46,7 +46,7 @@ public class ScheduledInterviewRequest {
   public String getIntervieweeId() {
     return intervieweeId;
   }
-  
+
   public String getRole() {
     return role;
   }
@@ -59,8 +59,8 @@ public class ScheduledInterviewRequest {
       ScheduledInterviewRequest that = (ScheduledInterviewRequest) o;
       return this.getId() == that.getId()
           && this.getDateString().equals(that.getDateString())
-          && this.getInterviewer().equals(that.getInterviewerId())
-          && this.getInterviewee().equals(that.getIntervieweeId())
+          && this.getInterviewerId().equals(that.getInterviewerId())
+          && this.getIntervieweeId().equals(that.getIntervieweeId())
           && this.getRole().equals(that.getRole());
     }
     return false;
@@ -68,7 +68,7 @@ public class ScheduledInterviewRequest {
 
   public String toString() {
     return String.format(
-        "{%s:%s, %s:%s, %s:%s, %s:%s}",
+        "{%s:%s, %s:%s, %s:%s, %s:%s, %s:%s}",
         "id",
         id,
         "dateString",
--- a/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
+++ b/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
@@ -17,6 +17,7 @@ package com.google.sps.data;
 /** Represents the data sent in a put or post request to the Person Servlet. */
 public class ScheduledInterviewRequest {
   private long id;
+<<<<<<< HEAD
   private TimeRange when;
   private String interviewerId;
   private String intervieweeId;
@@ -27,14 +28,26 @@ public class ScheduledInterviewRequest {
     this.when = when;
     this.interviewerId = interviewerId;
     this.intervieweeId = intervieweeId;
+=======
+  private String dateString;
+  private String interviewerEmail;
+  private String intervieweeEmail;
+
+  public ScheduledInterviewRequest(
+      long id, String dateString, String interviewerEmail, String intervieweeEmail) {
+    this.id = id;
+    this.dateString = dateString;
+    this.interviewerEmail = interviewerEmail;
+    this.intervieweeEmail = intervieweeEmail;
+>>>>>>> Formatted date is sent to page
   }
 
   public long getId() {
     return id;
   }
 
-  public TimeRange getWhen() {
-    return when;
+  public String getDateString() {
+    return dateString;
   }
 
   public String getInterviewerId() {
@@ -44,6 +57,7 @@ public class ScheduledInterviewRequest {
   public String getIntervieweeId() {
     return intervieweeId;
   }
+<<<<<<< HEAD
 
   public String toString() {
     return String.format(
@@ -64,4 +78,6 @@ public class ScheduledInterviewRequest {
     }
     return false;
   }
+=======
+>>>>>>> Formatted date is sent to page
 }
--- a/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
+++ b/src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
@@ -17,10 +17,10 @@ package com.google.sps.data;
 /** Represents the data sent in a put or post request to the Person Servlet. */
 public class ScheduledInterviewRequest {
   private long id;
-<<<<<<< HEAD
   private TimeRange when;
   private String interviewerId;
   private String intervieweeId;
+  private String role;
 
   public ScheduledInterviewRequest(
       long id, TimeRange when, String interviewerId, String intervieweeId) {
@@ -28,18 +28,7 @@ public class ScheduledInterviewRequest {
     this.when = when;
     this.interviewerId = interviewerId;
     this.intervieweeId = intervieweeId;
-=======
-  private String dateString;
-  private String interviewerEmail;
-  private String intervieweeEmail;
-
-  public ScheduledInterviewRequest(
-      long id, String dateString, String interviewerEmail, String intervieweeEmail) {
-    this.id = id;
-    this.dateString = dateString;
-    this.interviewerEmail = interviewerEmail;
-    this.intervieweeEmail = intervieweeEmail;
->>>>>>> Formatted date is sent to page
+    this.role = role;
   }
 
   public long getId() {
@@ -57,12 +46,9 @@ public class ScheduledInterviewRequest {
   public String getIntervieweeId() {
     return intervieweeId;
   }
-<<<<<<< HEAD
-
-  public String toString() {
-    return String.format(
-        "{%s:%s, %s:%s, %s:%s, %s:%s}",
-        "id", id, "when", when, "interviewerId", interviewerId, "intervieweeId", intervieweeId);
+  
+  public String getRole() {
+    return role;
   }
 
   public boolean equals(Object o) {
@@ -72,12 +58,26 @@ public class ScheduledInterviewRequest {
     if (o instanceof ScheduledInterviewRequest) {
       ScheduledInterviewRequest that = (ScheduledInterviewRequest) o;
       return this.getId() == that.getId()
-          && this.getWhen().equals(that.getWhen())
-          && this.getInterviewerId().equals(that.getInterviewerId())
-          && this.getIntervieweeId().equals(that.getIntervieweeId());
+          && this.getDateString().equals(that.getDateString())
+          && this.getInterviewer().equals(that.getInterviewerId())
+          && this.getInterviewee().equals(that.getIntervieweeId())
+          && this.getRole().equals(that.getRole());
     }
     return false;
   }
-=======
->>>>>>> Formatted date is sent to page
+
+  public String toString() {
+    return String.format(
+        "{%s:%s, %s:%s, %s:%s, %s:%s}",
+        "id",
+        id,
+        "dateString",
+        dateString,
+        "interviewerId",
+        interviewerId,
+        "intervieweeId",
+        intervieweeId,
+        "role",
+        role);
+  }
 }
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +112,62 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       return;
     }
   }
+
+  public List<ScheduledInterviewRequest> objectToRequestObject(
```
meagancheese: Pluralize this since it's acting on a list of objects. "objectsToRequestObjects".

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -17,53 +17,70 @@
...
       userId = String.format("%d", userEmail.hashCode());
     }
-    List<ScheduledInterview> scheduledInterviews = scheduledInterviewDao.getForPerson(userId);
-    response.setContentType("application/json;");
-    response.getWriter().println(new Gson().toJson(scheduledInterviews));
+    List<ScheduledInterviewRequest> scheduledInterviews =
+        objectToRequestObject(scheduledInterviewDao.getForPerson(userId), timeZoneId);
+    request.setAttribute("scheduledInterviews", scheduledInterviews);
+    List<ScheduledInterviewRequest> scheduledInterviews1 =
+        (List<ScheduledInterviewRequest>) request.getAttribute("scheduledInterviews");
```
meagancheese: I'm guessing this was for debugging, it should now be removed.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -17,53 +17,70 @@
...
+    RequestDispatcher rd = request.getRequestDispatcher("/scheduled-interviews.jsp");
+    try {
+      rd.forward(request, response);
+    } catch (ServletException e) {
+      e.printStackTrace();
+    }
```
meagancheese: I threw this Exception in the method statement when I encountered it, not sure which approach is best. I'm not sure when we would encounter this error or how we would test for it.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +112,62 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    String role;
+
+    if (userId.equals(scheduledInterview.interviewerId())) {
+      role = "Interviewer";
+    } else {
+      role = "Interviewee";
+    }
+
+    if (!personDao.get(scheduledInterview.interviewerId()).isPresent()) {
+      interviewer = "Nonexistent User";
```
meagancheese: I'm still not sure the "Nonexistent User" is a good option. We're generally assuming that interviews will only be scheduled with existing Persons and that a Person is never deleted.

src/main/webapp/scheduled-interviews.jsp
```diff
@@ -0,0 +1,30 @@
+<%@page import="java.util.ArrayList"%>
+<%@page import="java.util.List"%>
+<%@ page import="com.google.sps.data.ScheduledInterviewRequest" %>
+<%
+  List<ScheduledInterviewRequest> scheduledInterviews = (List<ScheduledInterviewRequest>)request.getAttribute("scheduledInterviews");
```
meagancheese: Space between ) and request.

src/main/webapp/scheduled-interviews.jsp
```diff
@@ -0,0 +1,30 @@
...
+  <c:otherwise>
+    <c:forEach items= "${pageScope.scheduledInterviews}" var="scheduledInterview">
+      <div class="row">
+        <div class="card w-75 scheduled-interview-card">
+          <div class="card-body">
+            <h5 class="card-title">Your Role: ${scheduledInterview.getRole()}</h5>
+            <p class="card-text">${scheduledInterview.getDateString()}</p>
+          </div>
+          <ul class="list-group list-group-flush">
+            <li class="list-group-item">Interviewee: ${scheduledInterview.getIntervieweeId()}</li>
```
meagancheese: This should not be an id, but rather a first name. You might need a separate class just for generating the jsp. (Replacing the id fields with names)

src/main/webapp/scheduled-interviews.jsp
```diff
@@ -0,0 +1,30 @@
...
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+
+<c:choose>
+  <c:when test= "${empty scheduledInterviews}">
+    <h2 style="text-align: center">No Scheduled Interviews</h2>
+  </c:when>
+  <c:otherwise>
+    <c:forEach items= "${pageScope.scheduledInterviews}" var="scheduledInterview">
+      <div class="row">
+        <div class="card w-75 scheduled-interview-card">
```
meagancheese, line 17: It might be worth-while to include the scheduled interview id somewhere (in a data-id field maybe) for easy access later on.

src/main/webapp/search-interview.js
```diff
@@ -40,8 +44,20 @@ function selectInterview(interviewer) {
       `PM with a ${interviewer.getAttribute('data-company')} ` +
       `${interviewer.getAttribute('data-job')}. Check your email for more ` +
       `information.`);
+    const request = new Request(`/scheduled-interviews?startTime=2020-07-13T18:30:10Z&endTime=2020-07-13T19:30:10Z&interviewer=${interviewer.getAttribute('data-email')}&interviewee=	120316124221414916520`, {
```
meagancheese: Too long of a line

meagancheese: If you're using this for testing, please don't, this is the code I'm working on rn. Directly access Datastore and add a scheduled interview to display it or else remove it now that you know it works.

meagancheese (changes requested):
Some nits

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/webapp/scheduled-interviews.jsp
GrantComm: Done

src/main/webapp/scheduled-interviews.jsp
GrantComm: Done, I changed the name of the methods in the scheduledInterviewRequest class. I use that class to get the names associated with the interviewer and interviewee ids and display them on the page.

src/main/webapp/search-interview.js
GrantComm: Done

src/main/webapp/scheduled-interviews.jsp
GrantComm: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: I'll ask during the sync

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +110,62 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       return;
     }
   }
+
+  public List<ScheduledInterviewRequest> objectToRequestObjects(
```
meagancheese: Maybe interviewsToRequestObjects?

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +110,62 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    return requestObjects;
+  }
+
+  private String getDateString(TimeRange when, ZoneId timeZoneId) {
+    LocalDateTime start = LocalDateTime.ofInstant(when.start(), timeZoneId);
+    LocalDateTime end = LocalDateTime.ofInstant(when.end(), timeZoneId);
+    String startTime = start.format(DateTimeFormatter.ofPattern("h:mm a"));
+    String endTime = end.format(DateTimeFormatter.ofPattern("h:mm a"));
+    String day = start.format(DateTimeFormatter.ofPattern("EEEE, MMMM d, yyyy"));
+    return day + " from " + startTime + " to " + endTime;
```
meagancheese: Use String.format() here instead of String concatenation.

src/main/webapp/search-interview.js
```diff
@@ -16,6 +16,10 @@ function onSearchInterviewLoad() {
   const loginInfo = getLoginInfo();
   loginInfo.then(supplyLogoutLinkOrRedirectHome); 
   loginInfo.then(getUserOrRedirectRegistration);
+  let email = loginInfo.then(getUserOrRedirectRegistration).then((person) => {
+    return person.email;      
+  }); 
+  console.log(email);
```
meagancheese: What is this for?

meagancheese (changes requested):
A few more things

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/webapp/search-interview.js
GrantComm: Shouldnt be there, removed it.

meagancheese (approved):
LGTM % exception question

meagancheese:
Also change title

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +110,62 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  private String getDateString(TimeRange when, ZoneId timeZoneId) {
+    LocalDateTime start = LocalDateTime.ofInstant(when.start(), timeZoneId);
+    LocalDateTime end = LocalDateTime.ofInstant(when.end(), timeZoneId);
+    String startTime = start.format(DateTimeFormatter.ofPattern("h:mm a"));
+    String endTime = end.format(DateTimeFormatter.ofPattern("h:mm a"));
+    String day = start.format(DateTimeFormatter.ofPattern("EEEE, MMMM d, yyyy"));
+    return String.format("%s from %s to %s", day, startTime, endTime);
+  }
+
+  private ScheduledInterviewRequest makeSheduledInterviewRequest(
```
apluscs: Nit: makeScheduledInterviewRequest
(the c) and change use cases to match

src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
```diff
@@ -17,38 +17,38 @@
...
-      long id, TimeRange when, String interviewerId, String intervieweeId) {
+      long id, String dateString, String interviewer, String interviewee, String role) {
     this.id = id;
-    this.when = when;
-    this.interviewerId = interviewerId;
-    this.intervieweeId = intervieweeId;
+    this.dateString = dateString;
+    this.interviewer = interviewer;
+    this.interviewee = interviewee;
+    this.role = role;
```
apluscs, line 31: Q: l thought we were using role for interviewee/interviewer and position for SWE/PM/.... ?

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +110,62 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      ScheduledInterview scheduledInterview, ZoneId timeZoneId) {
+    String userEmail = UserServiceFactory.getUserService().getCurrentUser().getEmail();
+    String userId = UserServiceFactory.getUserService().getCurrentUser().getUserId();
+    if (userId == null) {
```
apluscs, line 143: Q: when would userId ever be null?

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +110,62 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       return;
     }
   }
+
+  public List<ScheduledInterviewRequest> scheduledInterviewsToRequestObjects(
+      List<ScheduledInterview> scheduledInterviews, String timeZoneIdString) {
+    ZoneId timeZoneId;
+    if (timeZoneIdString == null) {
```
apluscs: Q: when would this ever be null?

apluscs (commented):
Mostly questions

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
meagancheese: This is because there isn't a good enough mock for .getUserId() to not be null. Solely used for testing, Grant maybe copy over a comment each time this is called. This is what I've been using:
// Since UserId does not have a valid Mock, if the id is null (as when testing), it will be
    // replaced with this hashcode.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: It was null for my tests but I just added a parameter in the tests and removed this.

src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
GrantComm: Not quite sure what the question is. The "role" here is what is displayed to the user who is participating in the interview. They are either an interviewer or interviewee depending on which id matches the user's id

apluscs (approved):

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
acmcarther: I discussed this with Grant a bit.

It seems unclear what else we ought to do in this case. This function isn't really responsible for fixing instances of ScheduledInterview whose associated Person(s) aren't in the DB. The only choice really is to either (1) render something somewhat weird or (2) throw an exception we do not catch and return a 500.

(1) seems better to me.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +114,59 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    String userId = UserServiceFactory.getUserService().getCurrentUser().getUserId();
+    // Since UserId does not have a valid Mock, if the id is null (as when testing), it will be
+    // replaced with this hashcode.
+    if (userId == null) {
+      userId = String.format("%d", userEmail.hashCode());
+    }
+    String date = getDateString(scheduledInterview.when(), timeZoneId);
+    String interviewer;
+    String interviewee;
+    String role;
```
acmcarther: I'm not really a fan of this forward declaration style. It opens the door to uninitialized variable issues. Try to write your variables so that they're (1) initialized immediately  (2) declared just when needed and (ideally) (3) immutable/const.

Following these rules reduces the amount of mental load for reviewers. Carrying large numbers of potentially initialized variables in your head can be painful.

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +114,59 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+
+    if (userId.equals(scheduledInterview.interviewerId())) {
+      role = "Interviewer";
+    } else {
+      role = "Interviewee";
+    }
+
+    if (!personDao.get(scheduledInterview.interviewerId()).isPresent()) {
```
acmcarther: There's an API for this.

```java
String interviewer = 
    personDao.get(scheduledInterview.interviewerId())
        .orElse("Nonexistent User");
```

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +114,59 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      interviewer = "Nonexistent User";
+    } else {
+      interviewer = personDao.get(scheduledInterview.interviewerId()).get().firstName();
+    }
+
+    if (!personDao.get(scheduledInterview.intervieweeId()).isPresent()) {
```
acmcarther: See above (on `Optional.orElse`)

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -95,4 +114,59 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    }
+    String date = getDateString(scheduledInterview.when(), timeZoneId);
+    String interviewer;
+    String interviewee;
+    String role;
+
+    if (userId.equals(scheduledInterview.interviewerId())) {
+      role = "Interviewer";
+    } else {
+      role = "Interviewee";
```
acmcarther: This isn't really a safe assumption. One of the most common sources of bugs is folks assuming false for one boolean (state) means true for another. Often a binary choice like that grows to more than two options, and then all this logic becomes broken. Best to handle the "unknown" case as soon as you write this logic.

To that end, I recommend you pull this out into a function.

```java
static String UserRole(ScheduledInterview scheduledInterview, String userId) {
  if (userId.equals(scheduledInterview.interviewerId()) {
    return "interviewer";
  }
  if (userId.equals(scheduledInterview.intervieweeId()) {
    return "interviewee";
  }
  return "unknown";
}
```

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
```diff
@@ -17,59 +17,78 @@
...
 
   // Gets the current user's email from request and returns the ScheduledInterviews for that person.
-  // If the email that is requested matches the email that is logged in, then the scheduled
-  // interviews are returned, otherwise SC_UNAUTHORIZED is returned.
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String timeZoneId = request.getParameter("timeZone");
     String userEmail = UserServiceFactory.getUserService().getCurrentUser().getEmail();
     String userId = UserServiceFactory.getUserService().getCurrentUser().getUserId();
+    // Since UserId does not have a valid Mock, if the id is null (as when testing), it will be
```
acmcarther, line 67: Copying multi line comments multiple times is a red flag. Code duplication is fine sometimes, but if each instance of duplication needs an explanation then that means you ought to be doing something else.

Create a helper somewhere (not here, probably), that gets the user id or a mock id.

acmcarther (changes requested):
first pass

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
GrantComm: Done

src/main/java/com/google/sps/servlets/ScheduledInterviewServlet.java
acmcarther: Disregard, discussed in standup

(this sucks, but I don't think we have time budget to fix it)

acmcarther (approved):

PR #118, opened 2020-07-22
Refactors all Person files to add an id field and avoid passing around the user's email.

meagancheese:


src/main/java/com/google/sps/servlets/PersonServlet.java
```diff
@@ -55,11 +55,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       response.sendError(400);
       return;
     }
-    if (!authenticateRequest(personRequest)) {
-      response.sendError(HttpServletResponse.SC_UNAUTHORIZED);
-      return;
-    }
-    personDao.update(Person.create(personRequest));
+    UserService userService = UserServiceFactory.getUserService();
```
acmcarther: Can this be initialized once for the whole servlet? It's probably nontrivial to construct.

src/main/java/com/google/sps/data/FakePersonDao.java
```diff
@@ -14,39 +14,58 @@
...
-    datastore = new HashMap<String, Person>();
+    storedObjects = new HashMap<Long, Person>();
   }
 
-  /** We put person into datastore with email as the key. */
+  /** We put person into storedObjects with a randomly generated long as its id. */
   @Override
   public void create(Person person) {
-    datastore.put(person.email(), person);
+    long id = new Random().nextLong();
```
acmcarther: I see that UserService's User type has as "getUserId".

https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User.html#getUserId--

Does it make sense to use this ID instead of our own ID? This would eliminate the need to look up the user by email on every call.

acmcarther (changes requested):
First pass, good work!

I found https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User.html#getUserId--, so I've asked about using that instead.

src/main/java/com/google/sps/servlets/PersonServlet.java
meagancheese: Done

src/main/java/com/google/sps/data/FakePersonDao.java
meagancheese: Done

apluscs (approved):
Sorry for the delay, LGTM.
Do you know what getUserId() looks like by any chance? ls it a messy hash or something similar to their gmail username?

GrantComm (approved):
LGTM

acmcarther (approved):

PR #119, opened 2020-07-23
Fixed a profile form verification bug

apluscs:
The profile form wasn't being verified before sending to backend. Now it is.

### DIFF
--- a/src/main/webapp/profile.html
+++ b/src/main/webapp/profile.html
@@ -41,7 +41,7 @@
       <br>
       <h1 class="text-center">Your Profile</h1>
       <form class="needs-validation" id="profile-form" novalidate 
-          onsubmit="event.preventDefault(); submitProfileForm('PUT','/profile.html')">
+          onsubmit="event.preventDefault(); submitProfileForm('PUT','/profile.html');">
         <div class="form-row">
           <div class="col-md-12 mb-3">
             <label for="user-email">Email</label>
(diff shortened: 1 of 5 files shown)


src/main/webapp/profile.js
```diff
@@ -40,6 +42,12 @@ function submitProfileForm(methodType, redirectUrl) {
     });
 }
 
+function validateProfileForm() {
```
meagancheese, line 45: Where is this called?

meagancheese (changes requested):
One question about the validation process.

src/main/webapp/profile.js
apluscs: image
When they submit on profile or registration page

meagancheese (approved):
LGTM

GrantComm (approved):
LGTM

src/main/webapp/profile.html
```diff
@@ -40,7 +40,8 @@
       </nav>
       <br>
       <h1 class="text-center">Your Profile</h1>
-      <form class="needs-validation" novalidate action="/person" method="POST">
+      <form class="needs-validation" id="profile-form" novalidate 
+          onsubmit="event.preventDefault(); submitProfileForm('PUT','/profile.html')">
```
jalexanderqed: Semicolon after `submitProfileForm()`.

jalexanderqed (approved):

PR #120, opened 2020-07-23
Displays registration prompt on home page if logged in but not registered.

apluscs:
image

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -16,6 +16,10 @@ function onIndexLoad() {
   checkLogin();
 }
 
+// Disables / enables navbar links and adjusts login / logout tab depending on if user is logged in 
+// or out.
+// Displays login message if user is not logged in and registration message if user is logged in 
+// but not registered.
 function checkLogin() {
   fetch('/login').then(response => response.json()).then(status => {
     if(status.loggedIn){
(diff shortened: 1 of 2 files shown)


GrantComm (approved):
LGTM

meagancheese (approved):
LGTM

src/main/webapp/script.js
```diff
@@ -36,3 +41,14 @@ function checkLogin() {
     document.getElementById('login-tab').href = status.changeLogInStatusURL;
   });
 }
+
+// If user is not registered, show the registration message.
+function ifUnregisteredShowRegistrationMessage() {
+  fetch('/person')
+    .then(response => {
+      // response.redirected means they are not registered yet.
```
acmcarther, line 49: This is a hack. You're making the request for person information, but not to get that information, but just to know if the backend wants to redirect us.

Is `status.loggedIn` (in checkLogin) equivalent?

apluscs: No, this is to check if they are registered, not if they are logged in.

acmcarther: How are you going to check that they are registered if they are not logged in? If they're logged in then they are registered.

Does "registered" mean something different from "made an account"?

meagancheese: They log in with their google account (because we're using google's api) and then we ask them to make an account for our site by registering. So registering is making an account, but you do not need an account for our site to log in.

acmcarther (commented):
"registered" in the common parlance for web applications means that they have an account. If for some reason people that have accounts need to register, then we've created a confusing model.

src/main/webapp/script.js
acmcarther: That should be fixed. We should not require a separate step from oauth (that's bad UX).

PR #131, opened 2020-07-28
Finishes up the Qualifications work on the Profile and Registration pages.

meagancheese:
Adds the other 7 jobs to the list on the profile page and puts all 9 on the registration page. Allows for the qualified jobs information from the page to be sent to the Person Servlet for PUT and POST Requests and processes the same information from the GET Request. Also makes some adjustments to the placeholder settings (not showing placeholders when displaying the profile page except for when updating). The fetch error is also handled.

Screenshot 2020-07-28 at 6 24 47 PM

Screenshot 2020-07-28 at 6 26 28 PM

### DIFF
--- a/src/main/webapp/profile.js
+++ b/src/main/webapp/profile.js
@@ -76,4 +76,9 @@ function makeEditable() {
   currentJob.setAttribute('hidden', true);
   const jobField = document.getElementById('job-field');
   jobField.removeAttribute('hidden');
+  
+  const checkboxes = document.getElementsByClassName('check-editable');
+  Array.from(checkboxes).forEach(function(checkbox) {
+    checkbox.removeAttribute('disabled');
+  });
 }
--- a/src/main/webapp/profile.js
+++ b/src/main/webapp/profile.js
@@ -54,6 +54,7 @@ function autofillForm(person) {
   document.getElementById('first-name-field').value = person.firstName;
   document.getElementById('last-name-field').value = person.lastName;
   document.getElementById('company-field').value = person.company;
+  document.getElementById('current-job').textContent = person.job;
   document.getElementById('job-field').value = person.job;
   document.getElementById('linkedin-field').value = person.linkedIn;    
 }
--- a/src/main/webapp/profile.js
+++ b/src/main/webapp/profile.js
@@ -27,8 +27,9 @@ function submitProfileForm(methodType, redirectUrl) {
   }
   document.getElementById('company-field').placeholder = '';
   document.getElementById('linkedin-field').placeholder = '';
+  let possibleQualifiedJobs = document.getElementsByClassName('qualification');
   let qualifiedJobs = [];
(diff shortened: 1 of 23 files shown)


GrantComm (approved):

src/main/webapp/profile.js
```diff
@@ -42,6 +51,20 @@ function submitProfileForm(methodType, redirectUrl) {
     });
 }
 
+function possibleQualifiedJobs() {
```
apluscs: Better just make a class and document.getByClassName()

src/main/webapp/profile.js
```diff
@@ -24,13 +24,22 @@ function onProfileLoad() {
...
+      qualifiedJobs.push(job.getAttribute('data-enum-name'));
+    }
+  }
   const personJson = {
     firstName: $('#first-name-field').val(),
     lastName: $('#last-name-field').val(),
     company: $('#company-field').val(),
     job: $('#job-field').val(),
-    linkedin: $('#linkedin-field').val()
+    linkedin: $('#linkedin-field').val(),
```
apluscs: l'm really confused as to why company and linkedIn are null sometimes. linkedin might be because in PersonRequest, the i is capital. l'm happy to do the debugging myself, but am curious as to what you discovered.

apluscs (changes requested):

src/main/webapp/profile.js
meagancheese: I was about to look into this. I think you're right and it is just the i being capitalized or not. If that's all then I'll remove all my null checks.

meagancheese: Fixed it. It was the i vs I.

src/main/webapp/profile.js
meagancheese: Done

apluscs (approved):

src/main/webapp/register.html
```diff
@@ -66,12 +66,13 @@ <h5 class="text-center lead">Fill in the following fields below.</h5>
...
           <div class="col-md-6 mb-3">
             <label for="company-field">Company</label>
-            <input type="text" class="form-control" id="company-field" placeholder="Google" 
+            <input type="text" class="form-control" id="company-field" placeholder="Company"
                 value="" name="company">
           </div>
           <div class="form-group col-md-6 mb-3">
             <label for="job-field">Job</label>
             <select class="form-control" id="job-field" name="job">
+              <option></option>
```
acmcarther, line 75: Why this blank option?

acmcarther (approved):
Bummer that we can't test any of this, oh well.

src/main/webapp/register.html
meagancheese: This is so they can choose to not select a job, say if they are looking to apply to their first job and don't currently have a title.

PR #132, opened 2020-07-29
Fixed small error in TimeRange test file

GrantComm:
Fixed time assigned to Instant in test file


apluscs (approved):

meagancheese (approved):

acmcarther:
Does this bug reflect an issue in the underlying tests? Can you elaborate what the implications of this change is?

GrantComm:
> Does this bug reflect an issue in the underlying tests? Can you elaborate what the implications of this change is?

The time that was wrong only impacts two tests: rangeDurationIsZero() and contains_nonContaining(). Both tests still have the expected result as the time used doesn't break anything

acmcarther (approved):

PR #134, opened 2020-07-29
Availability refresh

meagancheese:
Refreshes the Availability Page when updated instead of reloading the table to avoid user confusion.

### DIFF
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -64,5 +64,5 @@ function updateAvailability() {
   };
   let requestBody = JSON.stringify(requestObject);
   let request = new Request('/availability', {method: 'PUT', body: requestBody});
-  fetch(request).then(unused => {loadAvailabilityTable(availabilityTableDiv(), browserTimezoneOffset())});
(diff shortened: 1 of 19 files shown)


apluscs (approved):

GrantComm (approved):

src/main/webapp/availability.js
```diff
@@ -64,5 +64,5 @@ function updateAvailability() {
   };
   let requestBody = JSON.stringify(requestObject);
   let request = new Request('/availability', {method: 'PUT', body: requestBody});
-  fetch(request).then(unused => {loadAvailabilityTable(availabilityTableDiv(), browserTimezoneOffset())});
+  fetch(request).then(unused => {location.reload()});
```
jalexanderqed: This should catch and alert on an error.

jalexanderqed (changes requested):

src/main/webapp/availability.js
meagancheese: Done

jalexanderqed (approved):

PR #139, opened 2020-07-30
Added okShadow field to Person

apluscs:
Also updated all use cases
Next step is to incorporate this with querying scheduledInterviews


src/main/java/com/google/sps/data/PersonRequest.java
```diff
@@ -64,21 +66,14 @@ public String getLinkedIn() {
...
-            "company",
-            company,
-            "job",
-            job,
-            "linkedIn",
-            linkedIn,
-            "qualifiedJobs")
+            "PutPersonRequest= firstName:%s, lastName:%s, company:%s, job:%s, linkedIn:%s, okShadow: %b, qualifiedJobs: \n",
+            firstName, lastName, company, job, linkedIn, okShadow)
         + qualifiedJobsToString();
```
meagancheese: This is weird, is there a reason qualifiedJobsToString() can't be within String.format?

meagancheese: Also, jobs should be before shadow to be consistent throughout.

meagancheese (changes requested):
One main thing, then make sure to resolve conflicts with master.

src/main/java/com/google/sps/data/PersonRequest.java
apluscs: Just realized qualifiedJobs is printed nicely, you were right.

meagancheese (approved):

GrantComm (approved):

jalexanderqed (approved):

PR #141, opened 2020-07-30
Frontend: Added okShadow field to Profile / Reg pages

apluscs:
Screenshot 2020-07-31 at 9 21 41 AM
Screenshot 2020-07-31 at 9 22 14 AM

### DIFF
--- a/src/main/webapp/profile.html
+++ b/src/main/webapp/profile.html
@@ -115,6 +115,14 @@
                 placeholder="https://www.linkedin.com/" value="" name="linkedin">
           </div>
         </div>
+        <div class="form-check">
+          <input disabled class="form-check-input check-editable" type="checkbox" 
+              id="ok-shadow-check">
+          <label class="form-check-label" for="ok-shadow-check">
+            I am OK if another person shadows my interviews.
+          </label>
+        </div>
+        <br>
         <div class="form-group row d-flex justify-content-center">
           <button class="btn btn-primary" id="update-button" type="button" 
               onclick="submitProfileForm('PUT','/profile.html')" hidden="">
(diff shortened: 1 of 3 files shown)


src/main/webapp/register.html
```diff
@@ -160,6 +160,13 @@ <h5 class="text-center lead">Fill in the following fields below.</h5>
                 value="" name="linkedin">
           </div>
         </div>
+        <div class="form-check">
+          <input class="form-check-input" type="checkbox" id="ok-shadow-check">
+          <label class="form-check-label" for="ok-shadow-check">
+            I am OK if another person shadows the interviews I have.
```
meagancheese: Weird that this is worded differently than on the profile page.

meagancheese (changes requested):
Make wording consistent and add pics to description.

src/main/webapp/register.html
apluscs: Good catch

meagancheese (approved):

src/main/webapp/profile.html
```diff
@@ -161,6 +161,14 @@ <h1 class="text-center">Your Profile</h1>
                 placeholder="" value="" name="linkedin">
           </div>
         </div>
+        <div class="form-check">
+          <input disabled class="form-check-input check-editable" type="checkbox" 
+              id="ok-shadow-check">
+          <label class="form-check-label" for="ok-shadow-check">
+            I am OK if another person shadows my interviews.
```
GrantComm: Could this be "Allow another person to shadow my interviews" instead?

GrantComm (approved):
Just one question.

acmcarther (approved):

src/main/webapp/profile.html
apluscs: Good idea. Will change.

apluscs: Good idea, will change.

PR #143, opened 2020-07-30
Returns <Not specified> when a potential interviewer has a blank company or job field.

meagancheese:


GrantComm (approved):

apluscs (approved):

src/main/java/com/google/sps/data/PossibleInterviewer.java
```diff
@@ -24,6 +24,12 @@
   public abstract String job();
 
   public static PossibleInterviewer create(String company, String job) {
+    if (company.equals("")) {
```
jalexanderqed, line 30: Could we do this on the JavaScript side instead? I think adding the fake values in JavaScript is a bit dangerous in case this object was ever used for a different purpose (i.e. in a way that could cause the "<Not specified>" to be written back to DataStore).

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/PossibleInterviewer.java
meagancheese: This object is currently only used for generating the front end, what did you have in mind with javascript?

meagancheese: I'll change it so there's a 'could-be-not-specified' class or something and change the text with a javascript function call before the text is displayed.

jalexanderqed (approved):

PR #150, opened 2020-07-31
Added shadowId field to ScheduledInterview

apluscs:
Also changed ScheduledInterviewRequest (used in JSP page) and updated all use cases of both classes.


src/main/java/com/google/sps/data/ScheduledInterviewRequest.java
```diff
@@ -62,6 +65,10 @@ public boolean getHasStarted() {
     return hasStarted;
   }
 
+  public String getShadow() {
+    return shadow;
+  }
+
   public boolean equals(Object o) {
```
meagancheese, line 72: Can we get rid of this function? Or at least make sure you're using .equals() for Shadow because it's a String?

src/test/java/com/google/sps/data/DatastoreScheduledInterviewDaoTest.java
```diff
@@ -123,19 +130,68 @@ public void getsAllScheduledInterviews() {
...
     List<ScheduledInterview> expected = new ArrayList<ScheduledInterview>();
     expected.add(copyScheduledInterview1);
     expected.add(copyScheduledInterview2);
     Assert.assertEquals(expected, result);
   }
 
+  // Tests whether all scheduledInterviews for a particular user when they are interviewee are
+  // retrieved.
+  @Test
+  public void getForPersonAsInterviewee() {
```
meagancheese: I think this and the next test are kind of unneccesary and could be replaced with one test where the user fills all 3 roles in the interviews returned.

meagancheese (approved):
Some nits, generally looks good.

src/test/java/com/google/sps/data/DatastoreScheduledInterviewDaoTest.java
apluscs: Will replace

GrantComm (approved):
LGTM, just fix merge conflicts

src/main/webapp/scheduled-interviews.jsp
```diff
@@ -22,6 +22,8 @@
           <ul class="list-group list-group-flush">
             <li class="list-group-item">Interviewee: ${scheduledInterview.getInterviewee()}</li>
             <li class="list-group-item">Interviewer: ${scheduledInterview.getInterviewer()}</li>
+            <!-- TODO: make sure this looks right after we finish the shadow flow. -->
```
acmcarther: I don't know if Jack or I commented on this, but TODOs without attribution (either a bug/github issue or a person's username) attached are generally discouraged. This it to prevent the code getting populated with unactionable todos.

This isn't super important, so I don't necessarily recommend a change here. Just wanted to call out something we appeared to have missed.

src/test/java/com/google/sps/servlets/ScheduledInterviewServletTest.java
```diff
@@ -182,6 +182,17 @@
               Instant.parse("2020-07-20T13:30:00Z"), Instant.parse("2020-07-20T13:45:00Z")),
           /*id=*/ -1,
           /*scheduled=*/ false);
+  private final Person shadow =
+      Person.create(
+          emailToId("shadow@gmail.com"),
+          "shadow@gmail.com",
+          "Shadow_first_name",
+          "",
```
acmcarther: These empty strings would generally merit a parameter comment as well (as is done for okShadow).

acmcarther (changes requested):
Generally LG, two comments.

acmcarther (approved):

src/main/webapp/scheduled-interviews.jsp
apluscs: l agree that the TODOs should be more concrete than just comments floating in the code. l made an issue.

src/test/java/com/google/sps/servlets/ScheduledInterviewServletTest.java
apluscs: Changed these to non empty strings

acmcarther (approved):

PR #158, opened 2020-08-04
Makes “InterviewMe” a link to home

GrantComm:
Screenshot 2020-08-04 at 8 56 08 AM


apluscs (approved):

meagancheese (approved):

src/main/webapp/about.html
```diff
@@ -19,15 +19,14 @@
   <body onload="onAboutLoad()">
     <div id="content">
       <nav class="navbar navbar-expand-lg navbar-light bg-light">
-        <span class="navbar-brand mb-0 h1">InterviewMe</span>
+        <a href= "index.html"><span class="navbar-brand mb-0 h1">InterviewMe</span></a>
```
jalexanderqed: I think these should instead have the path "/", which will result in a cleaner URL and achieve the same effect.

Also, please remove the space after the equals sign.

jalexanderqed (changes requested):

src/main/webapp/about.html
GrantComm: Done

jalexanderqed (approved):

PR #167, opened 2020-08-05
Enabled drag-to-select on Availability page

apluscs:
event.preventDefaults() are there so that it doesn't highlight the cells with light blue as it does when you select text

### DIFF
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+let mousedown = false;
+
 function onAvailabilityLoad() {
   const loginInfo = getLoginInfo();
   loginInfo.then(supplyLogoutLinkOrRedirectHome); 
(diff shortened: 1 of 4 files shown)


src/main/webapp/availability.js
```diff
@@ -12,22 +12,36 @@
...
+let mousedown = false;
+
 function onAvailabilityLoad() {
   const loginInfo = getLoginInfo();
   loginInfo.then(supplyLogoutLinkOrRedirectHome); 
   loginInfo.then(getUserOrRedirectRegistration);
   loadAvailabilityTable(availabilityTableDiv(), browserTimezoneOffset());
 }
 
+function markMouseDown(tile) {
```
GrantComm: How is the tile passed to this function used?

GrantComm (approved):
One question

src/main/webapp/availability.js
meagancheese: This shouldn't actually be taking in tile as a parameter.

meagancheese (approved):
Grant caught something, be sure to remove it. Otherwise, looks good.

src/main/webapp/availabilityTable.jsp
```diff
@@ -26,12 +26,12 @@
     </tr>
   </thead>
   <tbody>
-    <!-- TODO: Allow clicking and scrolling over multiple slots to select them.-->
     <!-- TODO: Change page format so that it is vertically condensed.-->
     <c:forEach var = "i" begin = "0" end = "${pageScope.list.get(0).size() - 1}">
       <tr>
         <c:forEach items = "${pageScope.list}" var = "day">
-          <td onclick="toggleTile(this)" data-utc="${day.get(i).utcEncoding()}" 
+          <td onmouseenter="event.preventDefault(); toggleTile(this);"
```
jalexanderqed, line 33: I would need to test it, but it seems like this implementation would have the effect that if the user clicks on box A, then drags over B, C, and D, A will not be selected because the mouse was not down when A was entered.. Could you test that?

jalexanderqed (changes requested):

src/main/webapp/availabilityTable.jsp
apluscs: Good catch, fixing now

jalexanderqed (approved):

PR #171, opened 2020-08-06
Attempts to fix indentation on availabilityTable.jsp

meagancheese:
I'm not sure how the td tags should fall now that the choose sequence includes the tr tag. Any guidance would be appreciated.

### DIFF
--- a/src/main/webapp/availabilityTable.jsp
+++ b/src/main/webapp/availabilityTable.jsp
@@ -37,13 +37,13 @@
         </c:otherwise>
       </c:choose>
             <td>${pageScope.list.get(0).get(i).time()}</td>
-              <c:forEach items = "${pageScope.list}" var = "day">
-                <td onmouseenter="event.preventDefault(); toggleTile(this);"
-                    onmousedown="event.preventDefault(); markMouseDown(); toggleTile(this);"
-                    data-utc="${day.get(i).utcEncoding()}" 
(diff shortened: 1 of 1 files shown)


src/main/webapp/availabilityTable.jsp
```diff
@@ -37,13 +37,13 @@
...
-                    class="${day.get(i).getClassList()}">
-                </td>
-              </c:forEach>
+            <c:forEach items = "${pageScope.list}" var = "day">
+              <td onmouseenter="event.preventDefault(); toggleTile(this);"
+                  onmousedown="event.preventDefault(); markMouseDown(); toggleTile(this);"
+                  data-utc="${day.get(i).utcEncoding()}" 
+                  class="${day.get(i).getClassList()}">
+              </td>
+            </c:forEach>
```
apluscs: l thought these would go on the same level as `</c:choose>` (line 38). Not sure though.

apluscs (approved):

src/main/webapp/availabilityTable.jsp
meagancheese: It looks like the indentation doesn't effect the performance at all, so I'll see what Alex thinks makes the most sense. I'm just weirded out by the tr tags not lining up.

GrantComm (approved):

acmcarther (approved):

PR #176, opened 2020-08-07
Handles disappearing availability bug with group transaction.

meagancheese:


apluscs (approved):

GrantComm (approved):

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
```diff
@@ -118,11 +118,15 @@ public void deleteInRangeForUser(String userId, Instant minTime, Instant maxTime
       keyList.add(entity.getKey());
     }
     // This iterative deletion avoids XG transactions, which max out at 25 root entities.
+    // We include the individual deletion transactions within a larger group transaction
+    // to avoid replacing Availabilities before they are deleted.
+    Transaction groupTxn = datastore.beginTransaction();
     for (Key key : keyList) {
       Transaction txn = datastore.beginTransaction();
```
jalexanderqed, line 125: I think these internal transactions should be taken out? Unless you can point me to documentation on what nested transactions mean for Datastore.

jalexanderqed (changes requested):

src/main/java/com/google/sps/data/DatastoreAvailabilityDao.java
meagancheese: I'm not sure on this point. If the inner transactions were removed would the last deletion still fully complete before the transaction finishes? Basically I'm asking if it would still wait for the current deletion to fully finish before moving on to the next or exiting entirely.

jalexanderqed (approved):

